PR #14, opened 2023-09-06
allow paged data type in rows

ShikamaruNaraHachidaime:
Allow paged data type in rows, like a Paginator, LengthAwarePaginator or CursorPaginator.


robsontenorio:
@ShikamaruNaraHachidaime  thanks!

I am finishing some fixes before release it.

PR #45, opened 2023-09-19
Select: allow disable options

essamamdani:
Example
```php
<x-select label="Type" icon="o-user" :options="[
                            ['id' => 'trending', 'name' => 'Trending Content','disabled'=>true],
                            ['id' => 'latest', 'name' => 'Latest Content'],
                            ['id' => 'information', 'name' => 'Information Content'],
                        ]"
                            wire:model="type" />
```


robsontenorio:
@essamamdani  Thanks!

PR #52, opened 2023-09-30
Update MenuItem.php

localdisk:
hi @robsontenorio 

Changed the URL judgment of MenuItem's activate-by-route attribute to exact match.

In the current implementation, `Str::startsWith` is used, so `/tags` and `/tags/create` are judged to be the same, and the  `background-color` of  `activate-by-route` is set for both. Applicable.

This fix can successfully correct this behavior.
I would like you to merge it.

### DIFF
--- a/src/View/Components/MenuItem.php
+++ b/src/View/Components/MenuItem.php
@@ -31,7 +31,7 @@ class MenuItem extends Component
         $link = url($this->link ?? '');
         $route = url(Route::current()->uri());
 
-        return Str::startsWith($route, $link);
+        return $link == $route;
     }
 
     public function render(): View|Closure|string


robsontenorio:
@localdisk  This is a expected behavior.

The menu item link is `/users`. Then, on page you click on `/users/1` or make a search the changes the url for`/user?name=mary`. Even `/users/1/comments`.

So, we expect the correspondent menu keep selected because this is related to same feature/context.

localdisk:
I see. I'll think of another way. thank you.

robsontenorio:
@localdisk the current behavior keep menu selected for sub url levels. 

What is your use case ?

localdisk:
This PR will be closed

PR #112, opened 2023-11-05
Include Close on Outside Click for Modal Component (Enabled by Default)

iammuttaqi:
**Description:**

In this PR, we have made significant changes to the Modal component. The modal is now designed to close by default when a user clicks outside or ESC key.

**New Feature:**

- We have introduced a new `persistent` prop. By default, the modal can be closed by clicking outside or pressing the ESC key. If you want to prevent this default closure behavior, you can use the `persistent` prop.

**Usage Example:**

```blade
<x-modal wire:model="myModal" persistent>
```

**Note:** _The author is welcome to make further adjustments or improvements to this code as needed. Feel free to modify the code to better._

### DIFF
--- a/src/View/Components/Modal.php
+++ b/src/View/Components/Modal.php
@@ -13,7 +13,7 @@ class Modal extends Component
         public ?string $title = null,
         public ?string $subtitle = null,
         public ?bool $separator = false,
-        public ?bool $closeOutside = true,
+        public ?bool $persistent = false,
 
         // Slots
         public ?string $actions = null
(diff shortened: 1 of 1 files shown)


robsontenorio:
@iammuttaqi Thanks for this!

Are you able to rework your PR like this ?

- By default it ALWAYS close with ESC or CLICK OUTSIDE `<x-modal>`. 
- Otherwise if want to avoid this behavior use `<x- modal persistent >`

iammuttaqi:
@robsontenorio Thank you for your input. I've updated the PR.

robsontenorio:
It looks good. I will prepare docs before releasing.

PR #163, opened 2023-11-24
Tags: New component for managing multiple tags

iammuttaqi:
**Tags: New component for managing multiple tags**

Web capture_24-11-2023_22438_localhost

### DIFF
--- a/src/View/Components/Tags.php
+++ b/src/View/Components/Tags.php
@@ -41,6 +41,7 @@ class Tags extends Component
 
                 hasTag(tag) {
                     var tag = this.tags.find(e => {
+                        e = e.toString();
                         return e.toLowerCase() === tag.toLowerCase()
                     })
                     return tag != undefined
@@ -83,7 +84,9 @@ class Tags extends Component
                         <input class="input input-bordered input-ghost input-sm my-1 rounded transition-all focus:input-primary" id="{{ $uuid }}" placeholder="{{ $attributes->whereStartsWith('placeholder')->first() }}" type="text" x-model="tag" x-on:click.outside="clear()" x-on:keydown.enter.prevent="push()" x-on:keyup.prevent="if (event.key === ',') { push() }" />
                     </div>
 
-                    <x-mary-icon class="inline-block h-6 w-6 min-w-fit cursor-pointer stroke-current transition-all hover:text-red-500" name="o-x-mark" x-on:click="clearAll()" x-show="tags.length" />
+                    <div>
+                        <x-mary-icon class="inline-block h-6 w-6 min-w-fit cursor-pointer stroke-current transition-all hover:text-red-500" name="o-x-mark" x-on:click="clearAll()" x-show="tags.length" />
+                    </div>
                 </div>
 
                 @error($modelName())
--- a/src/View/Components/Tags.php
+++ b/src/View/Components/Tags.php
@@ -44,8 +44,7 @@ class Tags extends Component
 
                     push() {
                         if (this.tag != '' && this.tag != null && this.tag != undefined) {
-                            let tag = this.tag.trim()
-                            tag = this.tag.replace(/,/g, '');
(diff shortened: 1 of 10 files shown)


src/View/Components/Tags.php
```diff
@@ -0,0 +1,102 @@
...
+                <!-- LABEL -->
+                <label class="label label-text pt-0 font-semibold" for="{{ $uuid }}">{{ $label }}</label>
+
+                <!-- DIV AS INPUT -->
+                <div class="input input-primary relative flex h-fit w-full max-w-full place-items-center justify-between [outline:none!important]">
+                    <div class="py-1">
+                        <template :key="index" x-for="(tag, index) in tags">
+                            <div class="my-1 mr-2 inline-block cursor-pointer space-x-2 rounded bg-primary/20 px-2 py-1 transition-all hover:bg-primary/40">
+                                <span x-text="tag"></span>
+                                <svg class="inline-block h-4 w-4 stroke-current transition-all hover:text-red-500" fill="none" viewBox="0 0 24 24" x-on:click="remove(index)" xmlns="http://www.w3.org/2000/svg">
```
robsontenorio, line 75: You can use here the internal x-mary-icon for “x” icon

src/View/Components/Tags.php
```diff
@@ -0,0 +1,102 @@
...
+                                    <path d="M6 18L18 6M6 6l12 12" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"></path>
+                                </svg>
+                            </div>
+                        </template>
+
+                        <!-- INPUT -->
+                        <input class="input input-bordered input-ghost input-sm my-1 rounded transition-all focus:input-primary" id="{{ $uuid }}" placeholder="{{ $attributes->whereStartsWith('placeholder')->first() }}" type="text" x-model="tag" x-on:click.outside="clear()" x-on:keydown.enter.prevent="push()" x-on:keyup.prevent="if (event.key === ',') { push() }" />
+                    </div>
+
+                    <svg class="inline-block h-6 w-6 min-w-fit cursor-pointer stroke-current transition-all hover:text-red-500" fill="none" viewBox="0 0 24 24" x-on:click="clearAll()" xmlns="http://www.w3.org/2000/svg">
```
robsontenorio, line 85: Same here.

iammuttaqi:
@robsontenorio PR updated

robsontenorio:
@iammuttaqi 

When initial list contains a number it fails

```php
public array $tags = ['a', 11, 'c'];
```

image
image


When I click this area it clears the list. Are you able to put clear icon right side, just like Choices?

Captura de Tela 2023-11-27 às 13 01 52

iammuttaqi:
@robsontenorio I couldn't find any issues, and it seems like everything is working as expected...

iammuttaqi:
@robsontenorio updated something. please check.

robsontenorio:
@iammuttaqi What was the bug?

iammuttaqi:
@robsontenorio it was still taking spaces

image

davidcabcab:
Any chance to use a tags list declared into a blade view instead using wire:model ?

This would be useful when iterating on a parent component, for example table rows. 

In this situation, tags would be accessible via  @php directly in the blade view. 

Unfortunately, < x-tags > seems to only read tags from the binded livewire variable.

Capture d’écran 2024-02-24 à 22 22 13

robsontenorio:
The maryUI components are made for Livewire 3. Everything can be handled using Livewire 3.

For now I don't have plans to support components on plain blade views.

robsontenorio:
Here a tricky to inject “external variables” into scopes. 

```
@scope(“xxx”, $var1. $var2, $var3)
```

davidcabcab:
> And yes, `wire:model` is the way.
> 
> Here is another trick when you need to inject external vars into scopes
> 
> ```
> @scope('xxxx', $mainVar, $var2, $var3)
> ```

That's brilliant, it helps a lot thanks !

PR #196, opened 2023-12-10
Fix tailwind force, remove dynamic class names

fratsloos:
When using multiple buttons as group, using daisyUI's [Join Component](https://daisyui.com/components/join/) I found an issue where the last button doesn't have a rounded border.

Screenshot 2023-12-10 161110

After checking the code I came to the conclusion this was caused because of the hidden `span` that was added after the button. As last element in the Join Component the `span` got the rounded border.

As far as I could find (and suggested by the comment) this `span` in the `Button` and `Stat` component is only used for the builder, because the classes of the `span` are not anywhere else in the code. Tailwind advises againt the use of [dynamic class names](https://tailwindcss.com/docs/content-configuration#dynamic-class-names).

For the `Button` and `Stat` component this is a straight forward fix, because the correct classes can be added in `$tooltipPosition`. For the `Toast` component I saw no other option than to use a custom attribute.

The result with the button group is now as expected:

Screenshot 2023-12-10 173530

Example code:
```blade
<div class="join" role="group">
    <x-button label="Play" class="btn-warning join-item" icon="s-play" />
    <x-button label="Stop" class="btn-warning join-item" icon="s-stop" />
</div>
```

Also fixed the source for the `Button` component where the `class` attribute was added twice. Before and after:

Screenshot 2023-12-10 171252
Screenshot 2023-12-10 171333

### DIFF
--- a/src/View/Components/Button.php
+++ b/src/View/Components/Button.php
@@ -10,7 +10,7 @@ class Button extends Component
 {
     public string $uuid;
 
-    public string $tooltipPosition = 'tooltip-top';
+    public string $tooltipPosition = 'lg:tooltip-top';
 
     public function __construct(
         public ?string $label = null,
@@ -30,7 +30,7 @@ class Button extends Component
     ) {
         $this->uuid = md5(serialize($this));
         $this->tooltip = $this->tooltip ?? $this->tooltipLeft ?? $this->tooltipRight ?? $this->tooltipBottom;
-        $this->tooltipPosition = $this->tooltipLeft ? 'tooltip-left' : ($this->tooltipRight ? 'tooltip-right' : ($this->tooltipBottom ? 'tooltip-bottom' : 'tooltip-top'));
+        $this->tooltipPosition = $this->tooltipLeft ? 'lg:tooltip-left' : ($this->tooltipRight ? 'lg:tooltip-right' : ($this->tooltipBottom ? 'lg:tooltip-bottom' : 'lg:tooltip-top'));
     }
 
     public function spinnerTarget(): ?string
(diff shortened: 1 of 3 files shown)


robsontenorio:
@fratsloos 

Thanks ! I will review it ASAP. 

But, looks like there is a bug on daisyUI related to hidden elements 

https://github.com/saadeghi/daisyui/issues/2330

fratsloos:
Ah, I didn't see that issue in `saadeghi/daisyui`. It is the same bug, where a hidden element is not ignored in the join. But apart from having the same outcome, the reason why the bug is triggered is different. In the case of the issue in `saadeghi/daisyui`, this is due to elements that may or may not be visible based on responsiveness.

In the code that has been modified by this PR, it is because there is an unnecessary extra element that is hidden. That element only exists to have the class names for Tailwind in the code. And this can be solved by not using dynamic class names elsewhere in the code.

robsontenorio:
@fratsloos 

Got it! Thanks for this!

By the way `x-classes` was a nice trick.

fratsloos:
Great that it's merged an you have released the changes! Keep up the good work! 👍 

I saw you added missing classes in my PR. I just wanted to let you know that I didn't add those classes in the original commit because I saw those classes in other files. For example the added `alert` classes were already in the [Toast and Errors components](https://github.com/search?q=repo%3Arobsontenorio%2Fmary+alert&type=code). And the `toast-top` and `toast-end` were used in the [Toast trait](https://github.com/search?q=repo%3Arobsontenorio%2Fmary+toast-&type=code).

Because all the files in the package are scanned because of the addition of the path in `tailwind.config.js` I thought that would be enough.

PR #213, opened 2023-12-17
Fix for use of invalid HTML id attribute

fratsloos:
When working with the tabs it was not possible to get them to work unless I used the code from the example in the [documentation](https://mary-ui.com/docs/components/tabs).

The following message returned in the browser console:

```console
Uncaught DOMException: Document.querySelector: '#8192874ab96e164e8e6767302891308c' is not a valid selector
    target http://localhost:8000/livewire/livewire.js?id=8a199ab2:2854
    skipDuringClone http://localhost:8000/livewire/livewire.js?id=8a199ab2:1632
    getTarget http://localhost:8000/livewire/livewire.js?id=8a199ab2:2857
    <anonymous> http://localhost:8000/livewire/livewire.js?id=8a199ab2:2814
    flushHandlers http://localhost:8000/livewire/livewire.js?id=8a199ab2:1135
    stopDeferring http://localhost:8000/livewire/livewire.js?id=8a199ab2:1140
    deferHandlingDirectives http://localhost:8000/livewire/livewire.js?id=8a199ab2:1143
    initTree http://localhost:8000/livewire/livewire.js?id=8a199ab2:654
    start http://localhost:8000/livewire/livewire.js?id=8a199ab2:609
    start http://localhost:8000/livewire/livewire.js?id=8a199ab2:608
    start2 http://localhost:8000/livewire/livewire.js?id=8a199ab2:8283
    <anonymous> http://localhost:8000/livewire/livewire.js?id=8a199ab2:9401
    EventListener.handleEvent* http://localhost:8000/livewire/livewire.js?id=8a199ab2:9400
    <anonymous> http://localhost:8000/livewire/livewire.js?id=8a199ab2:9404
[l](http://localhost:8000/livewire/livewire.js?id=8a199ab2)
```

The error arises because the `id` of the element in some cases starts with a number. This is a result of using the md5 function to generate the UUID.

For a `label` + `input` it does not matter much if the `id` attribute starts with a number. If the `id` is used in, for example, a selector in JavaScript, it does matter.

> Technically, the value for an id attribute may contain any character, except [whitespace](https://developer.mozilla.org/en-US/docs/Glossary/Whitespace) characters. However, to avoid inadvertent errors, only [ASCII](https://developer.mozilla.org/en-US/docs/Glossary/ASCII) letters, digits, '_', and '-' should be used, **and the value for an id attribute should start with a letter**.

[Source](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id)

This PR ensures that the `id` attribute always starts with a letter. Not only with the tabs, but also with the other components where the uuid is placed in the html `id` attribute.

Use the following view to test. The second set of tabs doesn't work for me on the `main` branch, but it does on PR.
[fix-tabs.blade.zip](https://github.com/robsontenorio/mary/files/13695483/fix-tabs.blade.zip)

### DIFF
--- a/src/Traits/WithHtmlId.php
+++ b/src/Traits/WithHtmlId.php
@@ -2,6 +2,7 @@
 
 namespace Mary\Traits;
 
+use Exception;
 use Illuminate\Support\Str;
 
 trait WithHtmlId
@@ -10,6 +11,10 @@ trait WithHtmlId
 
     public function setHtmlId(string $prefix, ?string $value = null): void
     {
+        if (preg_match('/^[a-z]/i', $prefix) !== 1) {
+            throw new Exception('Prefix must start with a letter.');
+        }
+
         $this->htmlId = sprintf(
             '%s-%s',
             $prefix,
--- a/src/Traits/WithHtmlId.php
+++ b/src/Traits/WithHtmlId.php
@@ -6,9 +6,9 @@ use Illuminate\Support\Str;
 
 trait WithHtmlId
 {
-    public string $htmlId;
+    public string $htmlId = 'element';
 
-    public function setHtmlId(string $prefix, ?string $value = null)
+    public function setHtmlId(string $prefix, ?string $value = null): void
     {
         $this->htmlId = sprintf(
             '%s-%s',
--- a/src/Traits/WithHtmlId.php
+++ b/src/Traits/WithHtmlId.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Mary\Traits;
+
+use Illuminate\Support\Str;
+
+trait WithHtmlId
+{
+    public string $htmlId;
+
+    public function setHtmlId(string $prefix, ?string $value = null)
+    {
+        $this->htmlId = sprintf(
+            '%s-%s',
+            $prefix,
+            $value ?? $this->uuid ?? Str::uuid(),
+        );
+    }
+}
--- a/src/Traits/WithUuid.php
+++ b/src/Traits/WithUuid.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Mary\Traits;
+
+trait WithUuid
+{
+    public string $uuid;
+
+    public function setUuid(string $uuid)
+    {
+        $this->uuid = $uuid;
+    }
+}
--- a/src/Traits/WithUuid.php
+++ b/src/Traits/WithUuid.php
@@ -6,7 +6,7 @@ trait WithUuid
 {
     public string $uuid;
 
-    public function setUuid(string $uuid)
+    public function setUuid(string $uuid): void
     {
         $this->uuid = $uuid;
     }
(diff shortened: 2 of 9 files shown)


robsontenorio:
@fratsloos Amazing PR! Will check soon.

fratsloos:
I just thought of the `setHtmlId` method and my comment:
> This PR ensures that the id attribute always starts with a letter. 

It doesn't, yet. Because there is no check on the content of the param of `setHtmlId`. It is still possible to start the `id` attribute with a number, as long as the param is a string. Eg: `1foobar`. There should be a check on the first char of the value of the param.

I'll fix that later today.

robsontenorio:
@fratsloos 

I have been using **uuid** approach to make sure some `wire:key` are unique across components. For Livewire it won't matter if uuid starts with a number. 

So, I took advantage of this technique and placed same uuid on HTML **id** on same component.

Could we make it simpler?

- Only one trait to generate UUID (`WithUuid`)
- Just add a random letter in front of generated Uuid with a proper logic.
- So we can keep all blade parts untouched 

So, the only change on all components would be.

```
$this->setUuid(md5(serialize($this)));
```

fratsloos:
I used the traits because it was not used in every component. The traits are now only used when the `uuid` or html `id` are used. If you want to have it in each file your solution will work. My suggestion would be to use [`Str::uuid`](https://laravel.com/docs/10.x/strings#method-str-uuid) as it generates a real UUID. But this MD5 hash will work just as good for what it does.

If you want to have it available in each component, why not go for an abstract class which you extend from each component? Any other logic that is needed in more than one class can be moved to that class as wel.

Example:
```php
namespace Mary\View\Components;

use Illuminate\View\Component;

abstract class MaryComponent extends Component {
    protected string $uuid;

    public function __construct()
    {
        $this->uuid = 'mary-' . md5(serialize($this)));
        // or: $this->uuid = sprintf('mary-%s', Str::uuid());

        parent::__construct();
    }
}
```

And in a component:

```php
<?php

namespace Mary\View\Components;

class Tab extends MaryComponent
{
    public function __construct() {
        dd($this->uuid);
    }
}
?>
```

robsontenorio:
@fratsloos 

I like your idea to have a base classe. But, for now I want to keep it simple to se how it goes, once not all components uses uuid or any other fancy extra internal default properties. Are willing to make that changes proposed by me?

And ... we can't use  **Str::uuid()** because every time component re-renders it will generate a random uuid. And that is a problem for `wire:key`. So, we need to stick to **md5** that ensures the same string will be generate on re-render.

fratsloos:
> I like your idea to have a base classe. But, for now I want to keep it simple to se how it goes, once not all components uses uuid or any other fancy extra internal default properties. Are you willing to make that changes proposed by me?

To be clear, you want me to update the PR to use the code in [this comment](https://github.com/robsontenorio/mary/pull/213#issuecomment-1861837982), or to implement a base class with only the `uuid`?

Either way I think I have time for that next Friday.

> And ... we can't use **Str::uuid()** because every time component re-renders it will generate a random uuid. And that is a problem for `wire:key`. So, we need to stick to **md5** that ensures the same string will be generate on re-render.

Yes you are right, I didn't think about that.

fratsloos:
I'll close this PR, as I have made the changes in a different branch. See #230.

PR #214, opened 2023-12-17
Fix cross-platform compatibility for `copyStubs`

dormammuuuuu:
This PR will fix the issues with windows 

image

I've updated the `copyStubs` function to work smoothly on both Windows and Unix-based systems. Now, it cleverly uses PHP's own tools to handle files and folders, making sure everything runs without a hitch no matter the platform. Plus, I've added some smart checks to avoid any mix-ups with already existing directories.

### DIFF
--- a/src/Console/Commands/MaryInstallCommand.php
+++ b/src/Console/Commands/MaryInstallCommand.php
@@ -142,35 +142,47 @@ It will set up:
     {
         $this->info("Copying stubs...\n");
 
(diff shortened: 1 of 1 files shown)


robsontenorio:
@dormammuuuuu thats great! Will check soon.

robsontenorio:
Thank you!

dormammuuuuu:
> Thank you!

Thanks :)

PR #227, opened 2023-12-21
Ignore displaying asterisk when "required" is false

iammuttaqi:
Fixed a bug where the asterisk (*) was always shown, ignoring the "required" attribute boolean value.

## Before
```php
<x-input :required="false" label="Name" wire:model="name" />
```
image

## After
```php
<x-input :required="false" label="Name" wire:model="name" />
```
image

```php
<x-input :required="true" label="Name" wire:model="name" />
```
```php
<x-input required label="Name" wire:model="name" />
```
image

### DIFF
--- a/src/View/Components/Input.php
+++ b/src/View/Components/Input.php
@@ -55,7 +55,7 @@ class Input extends Component
                         <span>
                             {{ $label }}
 
-                            @if($attributes->has('required'))
+                            @if($attributes->has('required') && $attributes->get('required'))
                                 <span class="text-error">*</span>
                             @endif
                         </span>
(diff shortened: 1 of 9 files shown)


src/View/Components/Input.php
```diff
@@ -55,7 +55,7 @@ public function render(): View|Closure|string
                         <span>
                             {{ $label }}
 
-                            @if($attributes->has('required'))
+                            @if($attributes->has('required') && $attributes->get('required'))
```
robsontenorio, line 58: If you only keep `$attributes->get('required')` will work?

iammuttaqi: this actually works, yes

robsontenorio:
@iammuttaqi If I not asking too much, can you check other input components?

robsontenorio:
Thank you ❤️

PR #230, opened 2023-12-22
Force UUID to start with letter

fratsloos:
See discussion in #213

Updated the UUID in all components by prepending it with a string, to avoid html `id`'s that start with a number.


robsontenorio:
Thanks Elwin!

PR #254, opened 2024-01-08
Adding opacity 0 to checkbox in theme toggle

neapolis79:
HI i used the new compoent and sow that the checkbox under the theme icon was still visible, so i put an opacity-0 proprerty to the chckbox

### DIFF
--- a/src/View/Components/ThemeToggle.php
+++ b/src/View/Components/ThemeToggle.php
@@ -44,7 +44,7 @@ class ThemeToggle extends Component
                         @mary-toggle-theme.window="toggle()"
                         {{ $attributes->class("swap swap-rotate") }}
                     >
-                        <input type="checkbox" class="theme-controller" @click="toggle()" :value="theme" />
+                        <input type="checkbox" class="theme-controller opacity-0" @click="toggle()" :value="theme" />
                         <x-mary-icon x-ref="sun" name="o-sun" class="swap-on" />
                         <x-mary-icon x-ref="moon" name="o-moon" class="swap-off"  />
                     </label>


neapolis79:
Before themetoggle
After themetoggle_after

robsontenorio:
@neapolis79  Ow! that was quick!

This component relies on https://daisyui.com/components/theme-controller/

I think you have an outdated version of daisyUI. Can you upgrade it and let me know if it is ok?

neapolis79:
no i have daisyUI 4.5.0, the newest version. i just checked

robsontenorio:
Just used in a new app and I don't have this issue ...

Hum ... Maybe some strange custom css on your side?


Anyway I will merge it.

neapolis79:
hi @robsontenorio  i found out why in my configurations create the issue....in my tailwind.config.js i had

module.exports = {
  //...
  plugins: [require("@tailwindcss/typography"), require("@tailwindcss/forms"), require("daisyui")],
}

if i remove the forms plugin like this
module.exports = {
  //...
  plugins: [require("@tailwindcss/typography"),  require("daisyui")],
}
the problem is not present anymore....

only now i saw in some issues that the official form plugin is not compatible with daisyUI....sorry

robsontenorio:
There is no problem. That opacity class don't hurts.

neapolis79:
i installed mary-ui in an existent project initially developed with Jestream, so i had already form plugin e typography in my config.js....then i switched to mary ui because i like the approach in component, in this transition the form plugin remained in the config.js because it was not indicated in daisyUI documentantation that was incompatibile

Issue #262, opened 2024-01-10
[Bug]: sidebar menu item links active class set incorrectly


gottaloveit:
### MaryUI version

1.17

### daisyUI version

4.5.0

### Livewire version

3.3

### What browsers are affected?

Firefox, Chrome, Microsoft Edge

### What happened?

In the sidebar, I have 2 test links:
 
/owner
/owner/listings

Everything ok when I am on /owner however on /owner/listings, both have active class set

![Screenshot 2024-01-10 140637](https://github.com/robsontenorio/mary/assets/316801/07311942-9889-4ca3-a9ed-cfebabf9e9f1)

robsontenorio:
@gottaloveit 

Can I see the code for this menu? Specially the link `property`

gottaloveit:
I think the startsWith function is causing the issue

```php
                <x-menu-item title="Home" icon="o-home" link="{{ route('owner.dashboard') }}" />
                <x-menu-item title="Messages" icon="o-envelope" link="###" />
                <x-menu-item title="My RVs" icon="mdi.rv-truck" link="{{ route('owner.listing.index') }}" />
```

routes
```php
   Route::prefix('owner')->group(function() {
        Route::name('owner.')->group(function() {

            Route::get('/', function () {
                return view('pages.owner.dashboard');
            })->name('dashboard');

            Route::prefix('listings')->group(function () {
                Route::name('listing.')->group(function () {

                    Route::get('/', function () {
                        return view('pages.owner.listing-index');
                    })->name('index');

                    Route::get('/create', function () {
                        return view('pages.owner.listing-create');
                    })->name('create');

                    Route::get('/update/{listing}', function (Listing $listing) {
                        return view('pages.owner.listing-update', ['listing' => $listing]);
                    })->name('update');
                });
            });

        });
    });
```

so the code produces links like:
/owner
/owner/listings

gottaloveit:
shouldn't the check below be enough to set active class?  I mean, you want definite match on url and link right?  why have a fallback if it does not match completely, to in this case, match it on prefix ?

```php
if ($link == $route) {
            return true;
        }
```

robsontenorio:
@gottaloveit 



I suggest to use different base name for each menu. 

- `/owners`
-  `/listings` or `/owners-listings` ...
 

**Your PR will break current behavior.** 

<hr />
<hr />
<hr />

[...]

PR #263, opened 2024-01-10
fix active class on links showing up twice

gottaloveit:
fixes robsontenorio/mary#262


robsontenorio:
For now I am closing this, as I can't realize how to have equivalent feature for the "route approach".

Just do as you suggested "manually add a check on each one and add classes".

Issue #283, opened 2024-01-29
[Bug]: Crop buttons not rendering correctly when using Jetstream


Ceepster14:
### MaryUI version

1.20

### daisyUI version

4.6.0

### Livewire version

3.4

### What browsers are affected?

Firefox, Chrome, Safari, Microsoft Edge

### What happened?

When using the crop option on a file upload in a Laravel project with Jetstream installed, the crop and cancel buttons don't render correctly as they're using the x-button component rather than x-mary-button component, so the :label parameters don't get passed in.

The issue may also occur elsewhere, but changing lines 184 & 185 in the File.php component to:

```
<x-mary-button :label="$cropCancelText" @click="close()" />
<x-mary-button :label="$cropSaveText" class="btn-primary" @click="save()" ::disabled="processing" />
```
 
appears to resolve the issue.

PR #284, opened 2024-01-29
fix button

robsontenorio:
Fixes: #283 

Uses `x-mary-button`


Ceepster14:
Thanks for the speedy fix, much appreciated!

PR #290, opened 2024-02-08
Update DatePicker.php

dormammuuuuu:
This PR will fix the Date Picker issue if you are using this component outside of livewire.


robsontenorio:
Jerico, I really appreciate your contribution. 

But for now I wanna keep things simple and focus only on Livewire, to avoid conditionals and workarounds on source code. 

In that case you could create a copy from this component on your own project and hack it. 

Thanks !

Issue #286, opened 2024-02-01
Mary Table - Create option for controlling cell class


joaoclobocar:
It would be valuable if Table component give the programmer the ability of controling cell class.

It will be useful in the case of screen size controlling the visibility of a column.

For example, if we have this headers.

```php
    public $headers = [
        ['key' => 'name', 'label' => 'Name', 'class' => ''],
        ['key' => 'city', 'label' => 'City', 'class' => 'hidden sm:table-cell'],
        ['key' => 'address', 'label' => 'Address', 'class' => 'hidden lg:table-cell'],
    ];
```
Currently, by existence of 'hidden' string on header class, the corresponding cell will be hidden, even if the title is displayed.

My suggestion is to create a key on $headers array to give full control to the cell visibily and properties:

```php
    public $headers = [
        ['key' => 'name', 'label' => 'Name', 'class' => ''],
        ['key' => 'city', 'label' => 'City', 'class' => 'hidden sm:table-cell', 'cell_class' => 'hidden sm:table-cell', ],
        ['key' => 'address', 'label' => 'Address', 'class' => 'hidden lg:table-cell', 'cell_class' => 'hidden sm:table-cell'],
    ];
```

robsontenorio:
It is a good idea. I will work on this and solve #285 also.

joaoclobocar:
For controlling visibility, I think keeping the header class in cell class would be enough ('same_cell_class' => true, 'keep_header_class' => true).

May be an array of classes for cell decoration could be switched by the key name plus some suffix (address_class, address_switch, etc) and merged to final class. It could help to avoid some scoping.

```PHP
    public $headers = [
        ['key' => 'name', 'label' => 'Name', 'class' => ''],
        ['key' => 'city', 'label' => 'City', 'class' => 'hidden sm:table-cell', 'cell_class' => 'hidden sm:table-cell', ],
        ['key' => 'address', 'label' => 'Address', 'class' => 'hidden lg:table-cell', 'same_cell_class' => true,
          'cell_decoration' => ['font-bold text-red-500', 'italic text-green-500']
        ],
    ];

    public $users = [
        ['name' => 'Joao', 'city' => 'Sao Paulo', 'address' => 'Av. Um, 249', 'address_class' => 0], // address in bold, red
        ['name' => 'Renata', 'city' => 'Campinas', 'address' => 'Avenida Dois, 249', 'address_class' => 1], // address in italic, green
    ];
```

Something like that.

[...]

PR #300, opened 2024-02-14
Table: loop and decoration

### LINKED ISSUE #292: Access the $loop variable in the Table component
Want to be able to identify the first and last table rows so I can change the buttons in the actions for these etc.

Thank You

robsontenorio:
- Allow to access the `$loop` context.
- Add row and cell decoration properties.

How to test

```zsh
composer require robsontenorio/mary:dev-table-loopcontext
php artisan view:clear.
```

image

image

image

image

image

image

### DIFF
--- a/src/View/Components/Table.php
+++ b/src/View/Components/Table.php
@@ -13,6 +13,8 @@ class Table extends Component
 {
     public string $uuid;
 
+    public mixed $loop = null;
+
     public function __construct(
         public array $headers,
         public ArrayAccess|array $rows,
@@ -69,7 +71,7 @@ class Table extends Component
     // Handle header sort
     public function getSort(mixed $header): mixed
     {
-        if (!$this->isSortable($header)) {
+        if (! $this->isSortable($header)) {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


joaoclobocar:
Thanks, @robsontenorio .

If the header class contains 'hidden' the cell will be hidden. 

The visibility control should be an user option, as suggested in #286 .

Changing from 

```PHP
<td @class([$cellClasses($row, $header['key']), "p-0" => $link, "hidden" => Str::contains($header['class'] ?? '', 'hidden') ])>
```

to

```PHP
<td @class([$cellClasses($row, $header['key']), "p-0" => $link, "hidden" => ! isset($this->cellDecoration[$key]) && Str::contains($header['class'] ?? '', 'hidden') ])>
```

can solve without breaking the previous implementations.

robsontenorio:
@joaoclobocar 

Nice, I will take a look about this “responsive” feature. 

Did you have a chance to test this branch?
Is everything ok?  Does it work for your use case ?

joaoclobocar:
`composer require robsontenorio:dev-loopcontext
php artisan view:clear.
`

robsontenorio:
@joaoclobocar 

I have update the original comment with the right branch and docs with usage about responsive CSS classes. Please, give a try.

The main usage of **header classes** is about "responsive classes" or "width". So it will be applied to respective columns and it makes sense for me.

It less verbose than introducing new extra `$headers` params. 


So, I am willing to introduce this sightly break change.

joaoclobocar:
Thank you for enhancing mary-ui.

lucas-davi:
thanks for the support, this is amazing

PR #315, opened 2024-02-25
Range: New Component

dormammuuuuu:
**Range** slider is used to select a value by sliding a handle.

Usage:

```
   <x-range
      wire:model="dataRange" // Binds the component to a Livewire model.
      min="0" // Optional: Specifies the minimum value. Defaults to 0 if not provided.
      max="100" // Optional: Specifies the maximum value. Defaults to 100 if not provided.
      step="10" // Optional: Defines the step size between values.
      class="range-accent" // Allows customization of the slider's color or size.
      label="Select a price range" // Optional: Displays a label for the slider.
      hint="This is the range hint" // Optional: Provides additional information or guidance.
      required // Marks the slider as required for form submission.
   />
```

Available Sizes:
`range-xs` | `range-sm` | `range-md` (default) | `range-lg`

<hr />

image

image

### DIFF
--- a/src/MaryServiceProvider.php
+++ b/src/MaryServiceProvider.php
@@ -44,6 +44,7 @@ use Mary\View\Components\Nav;
 use Mary\View\Components\Progress;
 use Mary\View\Components\ProgressRadial;
 use Mary\View\Components\Radio;
+use Mary\View\Components\Range;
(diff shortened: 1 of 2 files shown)


robsontenorio:
Thank you!

PR #318, opened 2024-02-26
Correct components to use internal x-mary- components

jedjones-uk:
Hi, 

two minor tweaks to fix issues I found while using the library, thanks for the hard work!

Correct Spotlight.php component to use x-mary-model
Correct ImageLibrary.php component to use x-mary-button


robsontenorio:
Thanks!

PR #331, opened 2024-03-03
Add configurable spotlight url

jedjones-uk:
Extend the Spotlight.php component to accept a url parameter which will tell it where to search. The default remains the same.

This is to enable easier integration with existing search sources, and better flexibility for users.


jedjones-uk:
It lacks control over the route for things such as middlewares and domains, which makes it difficult to use with multi-tenant systems.

It also forced you have a route hidden else where there is not obvious to any more junior members of a team.

robsontenorio:
Fair. Thank you !

PR #336, opened 2024-03-06
add additional config settings and condition the routes

juhniorsantos:
With this modification, it is possible to reduce the number of routes if any route is not used.

Another benefit is having the autonomy to create routes according to the project.

### DIFF
--- a/config/mary.php
+++ b/config/mary.php
@@ -24,7 +24,9 @@ return [
     'components' => [
         'spotlight' => [
             'class' => 'App\Support\Spotlight',
-        ]
+        ],
+        'toogle-sidebar' => true,
+        'upload' => true
     ]
 ];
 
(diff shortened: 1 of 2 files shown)


robsontenorio:
I am trying to figure out how bad is to have this additional routes, even if you are not using it. Many packages has that routes and you have notice.

And, what do you mean with "autonomy to crate routes"?

I am trying here to avoid extra configs to care about and let things as simples as possible.

juhniorsantos:
if I have an upload route with a different middleware than the traditional ones? And not the case of using "web" middleware, but specific for uploads.

Also thinking about security, if I have access to a system that uses the package, I can send a file without any kind of validation and execute it on the server

robsontenorio:
This upload route `/mary/upload` are used **only** by Editor and Markdown components, **not for your regular files** when you use `<input type="file" />`

It uses the Laravel your project's  disk settings (local or s3). You are safe here.

PR #337, opened 2024-03-07
WIP: Sidebar: better `collapse`

robsontenorio:
Try another approach for better positioning `collapse` button


jedjones-uk:
Hi @robsontenorio,

Thanks for taking another look at this, your approach works for my use case. I no longer get the issue with content longer than the screen height.

The one thing I did notice while testing is in mobile, I can't close the menu without selecting a menu item. Clicking away from the sidebar does nothing.

robsontenorio:
What branch have you tried that not works on moble? 

Please, share your full app template here.

jedjones-uk:
Yes, It's works as expected now, I think we can go ahead with this. Thank you again :)

PR #345, opened 2024-03-19
Icon: add icon set config option

REJack:
Add support to adjust/replace the hardcoded `heroicon` inside the Icon component to allow dot-less naming for non heroicons

### DIFF
--- a/config/mary.php
+++ b/config/mary.php
@@ -24,7 +24,10 @@ return [
     'components' => [
         'spotlight' => [
             'class' => 'App\Support\Spotlight',
-        ]
-    ]
+        ],
+        'icon' => [
+            'set' => 'heroicon',
+        ],
+    ],
 ];
 
--- a/src/View/Components/Icon.php
+++ b/src/View/Components/Icon.php
@@ -22,7 +22,7 @@ class Icon extends Component
     {
         $name = Str::of($this->name);
 
-        return $name->contains('.') ? $name->replace('.', '-') : "heroicon-{$this->name}";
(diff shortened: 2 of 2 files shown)


robsontenorio:
In order to make you we won’t introduce BC in the future, could you please refactor it to follow the current  hierarchy ?

“components - icon - set”

Just like we did on Spotlight config.

REJack:
I've adjusted the config hierachy 😄

robsontenorio:
Thank you!

REJack:
Ok, may we can use that for v2 release or so 😄

REJack:
I have thought about the revert, would it be option too if I would extend all the components with config-able icons? that would be a break point for me too, i don't like the heroicons 😅 I would create a WIP/Draft PR so you can track my changes from that and we can discuss over them directly.

winavin:
First of all, thank you for such a good package.

For icons required by internal components we can create separate "svg" folder in "resources". in balde-ui-icons we can create custom icon sets based on folder. so all mary-ui components will use mary-icons set. and user can use default icon set whatever he want.

for example

```
class ToastException extends Exception
{
    protected string $icon = 'o-information-circle';  // - remove
    protected string $icon = config('icons.defaults.toast-exception.icon');  // + add
```
``` 
config.php

"icons" =>[
"default-set" => "heroicons",
"defaults"=>[
    "toast-exception" => "mary.information-circle",
   // other components
],
]
```

this will solve all problems.
I know this will take some time, but hope to see something like this in next version.

PR #348, opened 2024-03-20
MenuTitle: new component & MenuSeparator: remove title and set correct style 

REJack:
I've added a new component MenuTitle & removed the title from MenuSeparator.

I've removed your default uppercase class to comply with daisyUI default style and it's hard to remove it if you don't want it (like me).

Example Images from maryUI-Docs - Left (modified) / Right (current)
image image
image image

In my use-case it looks like that

Dropdown with a simple MenuTitle and a single MenuSeparator
image

P.S.: mary-ui.com change/PR is coming too ;)

### DIFF
--- a/src/MaryServiceProvider.php
+++ b/src/MaryServiceProvider.php
@@ -43,6 +43,7 @@ use Mary\View\Components\Menu;
 use Mary\View\Components\MenuItem;
 use Mary\View\Components\MenuSeparator;
 use Mary\View\Components\MenuSub;
+use Mary\View\Components\MenuTitle;
 use Mary\View\Components\Modal;
 use Mary\View\Components\Nav;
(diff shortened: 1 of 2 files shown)


robsontenorio:
@REJack 

Why not modify the current `x-menu-separator` to have better look and feel? As I can see it would be just the "divider" style.

It currently has all necessary attributes (optional), so we would not need a new component.

REJack:
I've updated my PR, with that change my docs PR is not needed too, now I've updated the style + we keep the old title icon feature but with usage of my new MenuTitle. 

What you think about that?

I would still prefer a MenuTitle component to only have a title without separator/divider 😄.

robsontenorio:
Agreed. Reverted to the first commit. Thanks!

PR #349, opened 2024-03-21
WIP - All flexible error bags

jedjones-uk:
Hi @robsontenorio,

This is a PR to apply my flexible error bag approach to all the components that would require it. While doing so I noticed some things, and would like to suggest some more improvements.

I think the biggest one worth considering would be to abstract the logic into a trait such as `HasErrors` that would provide the properties and methods in a standardised way.

Doing so you could also standardise the $omitError behaviour and probably the actual html template as is commonly used.

You'd have the occasional situation where your usage in a component wouldn't use the trait template, but 90% of cases would be covered.

Finally, on the modelName() method, I have seen two different instances:
`return $this->attributes->whereStartsWith('wire:model')->first();`
and
`return $this->attributes->wire('model')->value();`

If we can standardise on which version to use it will make things a bit cleaner, and i'd still implement it using the modelName() method in the trait, so it can be overloaded reliably and without change in the component.

Let me know what you think, I am happy to make the changes if you agree, and will add more detail to the docs to cover it.


robsontenorio:
Ok, let’s try the trait and standardize things as you suggested. Please, go a head and thanks in advance.

jedjones-uk:
Hi @robsontenorio 

I have extracted things into the Trait and applied it to all relevant components, I haven't updated the docs, will try to get to that tomorrow.

some notes to help with review:

In the trait you will notice I access attributes with `$this->attributes` rather than a public property, this is because a public property in a trait doesn't seem to get auto wired up by the Blade component. There might be a way to make it work, but I couldn't figure it out.

My template is recursive to support two related situations:
1. There is more than 1 validation failure to return
2. there is more than 1 item to validate (tags for example)
The current template will handle that perfectly fine, so you could even have 3 validation rules fail on 4 tags and see all those errors.

Attributes:
`error-bag` This is the name of the error bag you wan't to reference, you can use a normal `val` or `val.*` wildcards
`omit-error` If this is set to true, you will still get highlight of border, but no actual error text
`first-error-only` If this is set to true, you will only ever see the first error.
`error-class` An escape hatch to overwrite the css rules for the output line, but keep everything else

Defaults:
`error-bag`           => `$modelName()`
`omit-error`          => `false`
`first-error-only`   =>`false`
`error-class`        =>`'text-red-500 label-text-alt p-1'`

Using the template in the components:

You will see i've done it this way `{!! $errorTemplate($errors) !!}`. I didn't want to have to pass in `$errors` but it is by far the most reliable way to get that value as it's bound to the view but not available in the trait. The alternative is playing around trying to pull the error bag out of session or request object, and it was not reliable.

Let me know what you think! :)

robsontenorio:
I will you finished the changes so I can review and see if there is some issues. 

I remember in the past I was afraid to introduce traits for some reason….

jedjones-uk:
If we have an issue with the Trait approach, the code would easily work as an abstract class to extend from as it doesn't use anything like a constructor. Worst case, I can always just move the code back into the components themselves

robsontenorio:
I have started this library with a philosophy that all components must work in self isolation. 

So I have avoided traits or abstracted class until now. When I see other UI libraries' code base it is really hard to understand what is going on. 

So I  shameless repeat code like “uuid” or “model names”. Maybe just standardize names worth it , without further refactoring.

jedjones-uk:
Not a problem, I will find time tomorrow to un-abstract the code and keep just the improvements. I will keep this philosophy in mind with future changes as well :)

jedjones-uk:
@robsontenorio I have removed the trait and made each of the components stand alone again. I have also written some updated documentation here: https://github.com/robsontenorio/mary-ui.com/pull/34

robsontenorio:
Thanks!
This week I will tackle all PRs.

robsontenorio:
@jedjones-uk 

Tailwind won't compile this classes placed here. It is working on your machine because some local cache from previous build.

```php
public ?string $errorClass = 'text-red-500 label-text-alt p-1',
```

You can use a dirty approach like we did it here. Does not matter where you place it, it must exists to be compiled.

[                        x-classes="alert alert-success alert-warning alert-error alert-info top-10 right-10 toast toast-top toast-bottom toast-center toast-end toast-middle toast-start"](https://github.com/robsontenorio/mary/blob/5d52543f151036a56c97185e93d0fab5c1b4d74c/src/View/Components/Toast.php#L34)

robsontenorio:
Also have tried your example and validation does not work as expected here.

image

jedjones-uk:
@robsontenorio I have updated the components with x-classes for tailwind and I have fixed the documentation.

robsontenorio:
I don't get the point of `error-bag`. 

I have removed all `error-bag="xxx"` from components. Then, changed the validation and it works.


image

robsontenorio:
And your current approach with `\Illuminate\Support\Facades\Validator::make($this->state,  ... )` triggers the validation from above form, even I had clicked on your form.

image

jedjones-uk:
@robsontenorio 

Regarding the `error-bag` question, I had wanted to avoid changing the validation rules to use the `state.name` style because some of the rules were coming from internal packages that couldn't be easily changed. Allowing the flexibility on the component level side stepped those issues.

`error bag` is definitely the wrong name and clashes with laravels own terms. It's not about laravel error bags but rather the error field name. I can change it to `error-field` if you think that's better?

I can fix the issue on the docs about triggering on the other forms.

robsontenorio:
Ok, “error-field” looks better. Please, change it. 

And regarding another form validation be triggering I suspect we can do nothing, if you are “firing” validation from Laravel validator.

jedjones-uk:
Field has been updated to error-field now

robsontenorio:
@jedjones-uk  

There is no push in this PR about this change.

jedjones-uk:
@robsontenorio Fixed :)

robsontenorio:
Thanks!

PR #351, opened 2024-03-22
feat(choices): Allow passing arguments to search function

Luc-cpl:
## Description

This pull request introduces the ability to pass additional arguments to the search function used within the `x-choices` component.

**Existing code with the original search-function behavior continues to work seamlessly.**

## Current Behavior:

The `x-choices` component currently allows specifying a search function name only. This function is called with only the search value as the argument.

## Proposed Change:

This PR modifies the x-choices component to accept additional arguments to be passed to the search function. This enables developers to provide more context to the search.

Example:

```html
<x-choices
    label="My choices component"
    wire:model="selectedItems"
    :options="options"
    search-function="search('{{$myValue}}', {{$someId}})"
    no-result-text="Nothing found!"
    searchable
/>
```

This will allow the parent component to handle it as:

```php
public function search(string $value, int $someId, string $search)
{
    // Implement your search logic using $value, $someId and $search arguments
}
```

### DIFF
--- a/src/View/Components/Choices.php
+++ b/src/View/Components/Choices.php
@@ -172,7 +172,7 @@ class Choices extends Component
                                     return
                                 }
 
-                                @this.{{ $searchFunction }}(value)
+                                @this.{{ str_replace(')(', ', ', $searchFunction . '(value)') }}
                             },
                             dispatchChangeEvent(detail) {
                                 this.$refs.searchInput.dispatchEvent(new CustomEvent('change-selection', { bubbles: true, detail }))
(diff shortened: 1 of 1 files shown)


robsontenorio:
Nice! Will try soon.

robsontenorio:
@Luc-cpl I have tried it. And it works.

But, for better DX can you adjust to work like this?

```php
public function search(string $search = '', string $value = '', int $someId = 0)
{
    // Keep `$search` as first param, because it is the default
}
```

Luc-cpl:
@robsontenorio sure.

Perhaps we can change the line that adds the value:
```
@this.{{ str_replace(')(', ', ', $searchFunction . '(value)') }}
```

To some like this:
```
@this.{{ str_contains($searchFunction, '(')
  ? preg_replace('/\((.*?)\)/', '(value, $1)', $searchFunction)
  : $searchFunction . '(value)' 
}}
```

I'm not on my personal computer now but I can test it later today.

Issue #358, opened 2024-03-29
customize modal size


T3mmyKay:
There should be a feature for customizing modal size

igoralentyev:
I know only one way to do this - style with your own normal css, not a tailwind. Cause classes we can assign in mary component only applies to dialog element itself, not to the div.box-modal which is controlling modal's layout in daisyui.

robsontenorio:
That is correct. The PR #359 it is no good because CSS would be inject in two parts of the component.

robsontenorio:
Release.

PR #359, opened 2024-03-29
Custom modal size

RealZone22:
I've added a `class` attribute to add custom classes to the `modal-box` div (#358)

Usage for custom modal size:
```html
<x-modal wire:model="myModal1" class="w-11/12 max-w-5xl">
    <div class="mb-5">Press `ESC`, click outside or click `CANCEL` to close.</div>
    <x-button label="Cancel" @click="$wire.myModal1 = false" />
</x-modal>
```

### DIFF
--- a/src/View/Components/Modal.php
+++ b/src/View/Components/Modal.php
@@ -12,7 +12,7 @@ class Modal extends Component
         public ?string $id = '',
         public ?string $title = null,
         public ?string $subtitle = null,
-        public ?string $class = null,
+        public ?string $boxClass = null,
         public ?bool $separator = false,
         public ?bool $persistent = false,
 
@@ -39,7 +39,7 @@ class Modal extends Component
                         @endif
                     @endif
                 >
-                    <div class="modal-box {{ $class }}">
+                    <div class="modal-box {{ $boxClass }}">
                         @if($title)
                             <x-mary-header :title="$title" :subtitle="$subtitle" size="text-2xl" :separator="$separator" class="mb-5" />
                         @endif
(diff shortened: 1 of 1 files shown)


igoralentyev:
this isnt work as you want if feature to close when clicking outside used.

RealZone22:
If I understand correctly, the feature to close the modal by clicking outside doesn't seem to work when a custom width is applied. Is that the case?

robsontenorio:
The PR #359 it is no good because CSS would be inject in two parts of the component.

igoralentyev:
> If I understand correctly, the feature to close the modal by clicking outside doesn't seem to work when a custom width is applied. Is that the case?

Yeah! Because in this case your width style applies to a dialog elem. And what you really want is a box-modal class.

RealZone22:
> > If I understand correctly, the feature to close the modal by clicking outside doesn't seem to work when a custom width is applied. Is that the case?
> 
> Yeah! Because in this case your width style applies to a dialog elem. And what you really want is a box-modal class.

I changed it back to `box-class` so you can use it like this:
```blade
<x-modal wire:model="myModal1" box-class="w-11/12 max-w-5xl">
    <div class="mb-5">Press `ESC`, click outside or click `CANCEL` to close.</div>
    <x-button label="Cancel" @click="$wire.myModal1 = false" />
</x-modal>
```

igoralentyev:
Great, i will try later!

leonardyrj:
Any plans for releasing this functionality?

robsontenorio:
Thanks!

mohamad-supangat:
@robsontenorio can you give document for using this feature, iam searching about it and find this

robsontenorio:
Its done.

PR #364, opened 2024-04-04
Menu item route param, default theme based on system preferences

dgtvc:
Hi!

I added default theme based on system preferences on alpine present plugin.
Don't why there is 2 params as $light and $dark that i don't see used anywhere.

For route param, this simply fixes my use case, where i have whole user panel under same prefix, and it makes few menu items active at the same time, simply using laravel based request()->routeIs() with * sign you can make only few route available.

Third thing is that i changed classes for sidebar to make it without scroll on sidebar. That the only thing that I'm not 100% sure of. But from my testing everything works great, even with bigger menus.


robsontenorio:
Thanks !

But you are trying to solve different things on same PR. 

Please, send 1 MR for each issue. Include some code example of the issue you are trying to solve.

robsontenorio:
Thank you!

PR #365, opened 2024-04-04
Colorpicker component

dgtvc:
To make it work, we need to add this code to head:

```
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@simonwep/pickr/dist/themes/nano.min.css"/>
<script src="https://cdn.jsdelivr.net/npm/@simonwep/pickr/dist/pickr.min.js"></script>
```

Screenshot:
image

### DIFF
--- a/src/View/Components/ColorPicker.php
+++ b/src/View/Components/ColorPicker.php
@@ -1,0 +1,177 @@
+<?php
+
+namespace Mary\View\Components;
+
+use Closure;
+use Illuminate\Contracts\View\View;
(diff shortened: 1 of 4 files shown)


dgtvc:
I just saw that i used wrong base commit and merged the second PR into this :( Will fix that later today

robsontenorio:
Please,

- Remove wrong commit 
- I think it’s better to use native device color picker instead of third party lib, because it is dead simple.

robsontenorio:
Solved by #388

PR #400, opened 2024-04-26
Alert: make alert dismissable

derrok:
This PR adds the ability to make alerts dismissable.

### DIFF
--- a/src/View/Components/Alert.php
+++ b/src/View/Components/Alert.php
@@ -15,6 +15,7 @@ class Alert extends Component
         public ?string $icon = null,
         public ?string $description = null,
         public ?bool $shadow = false,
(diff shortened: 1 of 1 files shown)


derrok:
This commit adds the default position for the toast. Now, you don't need to write
 `$this->success('All done!', 'toast-bottom toast-end')` 
each time in each of your components if you want to change the default position of your toast. You can write 
`<x-toast position="toast-bottom toast-end" />`
 in your layout and use the toast with the simple `$this->success('All done!')` command.

robsontenorio:
Thanks ! Looks good. 

On the next opportunity , please send one PR per feature. Because it’s easier to rollback in case of problems. 

But I will handle this as is.

PR #404, opened 2024-04-28
feat: add `sunday-start` modifier

itsmenewbie03:
This allows a cleaner way to make the calendar start on Sunday.

Currently to display make the calendar start on Sunday you have to do this
```blade
<x-mary-calendar locale="en-PH" :events="$events" weekend-highlight :config="['settings' => ['iso8601' => false]]" />
```
which IMO pretty verbose for a simple modification.


robsontenorio:
Thanks!

PR #405, opened 2024-04-28
Add thumbnail for articles/blog post to spotlight

Matildevoldsen:


robsontenorio:
Could you describe how it works ?

robsontenorio:
We already have the “avatar” option that is rounded full , but do you prefer it squared ?

It is easier to stick to “avatar” and override the style at “app.css” , than to duplicate the component code just to have a squared image.

PR #413, opened 2024-05-03
feat: Added w-full to Tabs

wishborn:
added the w-full class to the Tabs Alpine div to force tabs to take up the full pace of the container they are placed in.


robsontenorio:
Thanks!

PR #417, opened 2024-05-06
feat: improve inline labels for DatePicker and TextArea (#409)

### LINKED ISSUE #409: QUESTION: inline labels for DatePicker and TextArea
The inline labels for most input looks gorgeous, but for some input like the TextArea and DatePicker, they look a bit off
image
is this intentional?

itsmenewbie03:
Attempt to resolve #409 

## NOTE:
I don't see myself as a good UI/UX designer that's why I do love using this library. So I would highly appreciate inputs from experts in this field.

### DIFF
--- a/src/View/Components/Tabs.php
+++ b/src/View/Components/Tabs.php
@@ -36,7 +36,7 @@ class Tabs extends Component
                                      });
                                  }
                         }"
-                        class="relative"
+                        class="relative w-full"
                     >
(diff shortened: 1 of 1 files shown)


itsmenewbie03:
for some reason despite moving to another branch e99301a still got included in this PR, making it kinda a duplicate of #404

robsontenorio:
Thanks!

PR #432, opened 2024-05-11
Signature: Add initial image data on init

Karadox:
Populate the signature field with existing image data

### DIFF
--- a/src/View/Components/Signature.php
+++ b/src/View/Components/Signature.php
@@ -55,6 +55,9 @@ class Signature extends Component
                                 let canvas = document.getElementById('{{ $uuid }}signature')
                                 this.signature = new SignaturePad(canvas, {{ $setup() }});
 
+                                // Populate 
+                                this.signature.fromDataURL(this.value); 
+            
                                 // Resize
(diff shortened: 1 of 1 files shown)


robsontenorio:
Could you provide a example ?

Karadox:
Say you stored the result of the signature field in the database or you made a file. But you want to update it.
Now, just put the value into the modelfield te signature field uses and voila it shows the image as it was saved.

the value would be similar to what you would put as source for an  tag:

```
<?php
$this->signature = $storedValue;  // "data:image/png;base64,iVBORw0KGgoAAAANSUh.............QEWE="
?>

<x-signature wire:model="signature" />
```

robsontenorio:
This breaks the current usage.


```php
public string $signature1 = '';
```


```blade
<x-signature wire:model="signature1" hint="Please, sign it." />
```


image



And, thinking  about real world usage I am not seeing "real" benefit about this approach. Because once someone signed it and save it, you should not display the signature pad again, but a just an image.

If there is need to edit it, just show a empty signature pad again.

Issue #426, opened 2024-05-09
Choices component 'X' reset button still appears when no user is selected


ignaaaam:
### maryUI version

1.20

### daisyUI version

4.10.2

### Livewire version

3.4

### What browsers are affected?

Firefox, Chrome, Microsoft Edge

### What happened?

✅ For sure you can paste screenshots.

❌ Please, don't paste code as screenshot. Paste as formatted markdown code.

PHP

```php
public bool $modal = false;
```

HTML/Blade

```blade
<x-button label="Hi" />
```
Hello all, maybe this is intended, but on the Choices component, if I dont have any user or users selected the 'X' of reset still appears, shouldn't be that it doesnt appears when there's no user selected and appear once there's one or atleast one (if it's multiple)? Here's the two examples (blurred the users selected because it's name of real people and can't share that due to privacy)

**'X' appearing when no users selected**
<img width="624" alt="choices maryUI" src="https://github.com/robsontenorio/mary/assets/39315431/17fd6517-62c3-46bf-b721-31242f7b9b1f">

**'X' when users are selected** (thats how it should work when atleast one user is selected should appear instead of being present all time even is nothing is selected)

<img width="634" alt="choices mary users selected with blur" src="https://github.com/robsontenorio/mary/assets/39315431/e6021f20-a170-4036-8d64-be07868a380b">

PR #435, opened 2024-05-15
feat: hide 'X' button if no selection is made

itsmenewbie03:
This utilizes the existing getter `isSelectionEmpty` and tweaking it a bit to treat an empty string as empty. The hiding of the 'X' button is done with the x-show attribute.

implements #426


robsontenorio:
Thanks!

Just added same logic to ChoicesOffline.

PR #448, opened 2024-05-28
feat: add option close on kewydown escape

felipe-silveira-dev:
This allow close drawer on keydown escape


robsontenorio:
Thanks!

PR #453, opened 2024-05-29
Update Collapse.php

bestmomo:
Add `is-open` option


robsontenorio:
What this PR solves?

Currently you define the state like this. So, it renders as open.

```
public  $item = true;

<x-collapse wire:model="item" />
```

bestmomo:
I just wanted a shorter syntax for this option without creating a special property. But I can manage a property ;)

PR #457, opened 2024-05-30
Set names and allow prefix routes

prof-anderson-trindade:
Hello,

I have a specific case where I need to prefix all my routes. That's the reason I'm proposing those changes, including naming the routes.


robsontenorio:
Thanks!

Issue #483, opened 2024-06-13
Dropdown not working within a mary table cell


Enrica-r:
### maryUI version

1.31.0

### daisyUI version

4.11.1

### Livewire version

3.5.0

### What browsers are affected?

Chrome, Microsoft Edge

### What happened?

I use Mary table. In an action cell I want to use a dropdown component from Mary. I have two issues:

1. The UL is clipped like overflow-hidden
2. The UL opens on top of the trigger

The following picture shows how the dropdown opens. Only the last line of 4 list entries are showed:
![Dropdown within table cell UL clipped](https://github.com/robsontenorio/mary/assets/2463365/cd128594-0fd8-41bf-987a-06d06d03ce6e)

The next picture shows the exact same built component outside of the table:
![Dropdown outside table cell](https://github.com/robsontenorio/mary/assets/2463365/8596fdc5-d197-434f-ace0-081c7397b988)

The behaviour is caused by the class `overflow-x-auto` in the top div (with x-data) within table component. If I remove it the dropdown works but x-scrolling doesn't work anymore. To set the `overflow-y` to another value doesn't help.

Thank you for helping.

bestmomo:
I got it working like this:

```
<?php

namespace Mary\View\Components;

use Closure;
use Illuminate\Contracts\View\View;
use Illuminate\View\Component;

class Dropdown extends Component
{
    public string $uuid;

    public function __construct(
        public ?string $label = null,
        public ?string $icon = 'o-chevron-down',
        public ?bool $right = false,
        public mixed $trigger = null
    ) {
        $this->uuid = "mary" . md5(serialize($this));
    }

    public function render(): View|Closure|string
    {
        return <<<'HTML'
            <div x-data="{ open: false }" class="relative dropdown">
                <!-- CUSTOM TRIGGER -->
                @if($trigger)
                    <div x-ref="dropdownButton" @click="open = !open" {{ $trigger->attributes->class(['cursor-pointer']) }}>
                        {{ $trigger }}
                    </div>
                @else
                    <div x-ref="dropdownButton" @click="open = !open" {{ $attributes->class(["btn normal-case cursor-pointer"]) }}>
                        {{ $label }}
                        <x-mary-icon :name="$icon" />
                    </div>
                @endif

                <template x-if="open">
                    <div
                        class="fixed inset-0 z-50 items-center justify-center"
                        @click="open = false"
                    >
                        <div
                            x-show="open"
[...]

PR #485, opened 2024-06-13
Update Dropdown.php

bestmomo:


Enrica-r:
@bestmomo: Thank you very much for your quick solution.

The code line `:style="{'top': ($refs.dropdownButton.getBoundingClientRect().bottom + window.scrollY) + 'px', 'left':` has one disadvantage. If the trigger button is near the screen bottom, the box falls out of view.

I tried to replace it with `x-anchor.{{ $right ? 'bottom-end' : 'bottom-start' }}="$refs.dropdownButton"` but together with your class definition. Then it works as it should. Left and right positioning is working also.

bestmomo:
Yes there is this issue and another one, when screen is hight it no longer sticks to the trigger. This issue is hard to manage...

robsontenorio:
I think this issue can be solved by change some css class on original code from this component. 

I will give a try on next days.

robsontenorio:
Closing this, please follow #483

PR #489, opened 2024-06-15
Form : Add $noseparator

Jipem:
Default : $noseparator false (as now). 
If true, does not display hr above actions.


robsontenorio:
Thanks !

PR #490, opened 2024-06-15
Select : allow Enums

ryanvelbon:
Currently `<x-select>` seems to only work with simple arrays.

If we tried passing in an Enum

```html
@php
    $statusOptions = \App\Enums\PostStatus::cases();
@endphp

<x-select label="Status" :options="$statusOptions" wire:model="status" />
```

We would get the following error

```
Cannot use object of type App\Enums\PostStatus as array
```

### DIFF
--- a/src/View/Components/Select.php
+++ b/src/View/Components/Select.php
@@ -103,7 +103,11 @@ class Select extends Component
                         @endif
 
                         @foreach ($options as $option)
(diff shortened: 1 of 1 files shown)


robsontenorio:
Could share that enum definition?

robsontenorio:
Just modified to a simpler approach. Thanks!

ryanvelbon:
This is a way cleaner approach! Thank you for merging.

PR #491, opened 2024-06-15
Fix ImageLibrary block on crop error

ades4827:
When an error occurs while saving the crop the progress is not resetted so the component becomes unusable. By entering this small piece of code the error is displayed and you can continue using the component. To replicate the mistake: trying a crop on a large image with a size close to the upload limit threshold (the crop library produces a blob with a larger size than the original file)


robsontenorio:
And how resetting the progress will help this issue?
Could you please attach a video?

ades4827:
https://github.com/robsontenorio/mary/assets/27701869/391d1821-5da5-41fb-acb6-f118fc0305cc

as you can see after the error you can no longer click on the component because it is still "in progress". With my modification the component is unlocked and I can delete the image that is not right

robsontenorio:
Thanks !

PR #498, opened 2024-06-17
Add shadow on Statistic

bestmomo:
Capture d'écran 2024-06-17 151827


robsontenorio:
You can get this by using, for example,  `<x-stat class="shadow-md" .../>`.

That is the best approach instead of having extra parameters. At the past, I have added this extra attribute on `<x-card />` but I will deprecate it in the future.

robsontenorio:
Also a “mistake” from the very beginner version.

PR #499, opened 2024-06-17
MenuSub: add 'open' parameter

bimsonz:
Allow sub menus to default to an 'open' state by providing an optional parameter to the component constructor.


robsontenorio:
Thanks!

PR #508, opened 2024-06-25
Add 'disabled' option to Radio component options

bimsonz:
Allow an option set to 'disable' an option by defining `disabled === true`

### DIFF
--- a/src/View/Components/Radio.php
+++ b/src/View/Components/Radio.php
@@ -9,7 +9,6 @@ use Illuminate\View\Component;
 
 class Radio extends Component
 {
-
     public string $uuid;
 
     public function __construct(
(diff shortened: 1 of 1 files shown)


bimsonz:
Considered adding the ability to define the `disabled` key but it felt a little unnecessary... what do you think ?

robsontenorio:
Have you tested with this syntax `:disabled=“$myVar”` ?

bimsonz:
To be honest, im not that familiar with that syntax, is it used with Alpine ? I've not tested anything outside of the blade implementation. Happy to if you can point me in that direction

robsontenorio:
Any blade component variable can be bound to external variables . Try it `$myVar` as true/false and see.

bimsonz:
Is this suggestion to cater for disabling the whole radio component? 
My intension is to add the ability for specific options to be disabled.

Sorry, excuse my confusion here!

robsontenorio:
Thanks!

PR #515, opened 2024-07-06
DropdownComponent add new parameter "no-x-anchor" and "top"

s-kitade:
When there are a lot of Dropdown components side by side, the performance is very slow due to x-anchor

For this reason, we added the "no-x-anchor" property.

However, since there are cases where the menu expanded on the screen is filled up, we also added the "top" property.

```php
<x-dropdown no-x-anchor top>
....
```


robsontenorio:
Thanks!

PR #530, opened 2024-07-22
Add hidden column option

rezawikan:
It is possible to add an option to hide columns instead of deleting values ​​in the array headers at certain indexes and returning them back to the original order.


robsontenorio:
Thanks!

PR #532, opened 2024-07-23
Add support for Livewire model binding in Drawer component

r2luna:
Error: 
CleanShot 2024-07-23 at 10 41 33@2x

### DIFF
--- a/src/View/Components/Drawer.php
+++ b/src/View/Components/Drawer.php
@@ -5,6 +5,7 @@ namespace Mary\View\Components;
 use Closure;
 use Illuminate\Contracts\View\View;
 use Illuminate\View\Component;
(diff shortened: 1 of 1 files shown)


robsontenorio:
Hey , actually it already works with “wire:model”. 

What use case this PR fixes ?

Take a look on docs 
https://mary-ui.com/docs/components/drawer

r2luna:
It just a type return issue. 

When you have strict_types enable on the project it will get that error that I sent. 

Meaning that the $this->attribute->wire() returns an object not a string.

robsontenorio:
Thanks for these MR!

Can I be in touch with you? I am on twitter @robsontenorio

PR #542, opened 2024-07-28
Textarea: fix duplicated html id when using nested models.

anfeichtinger:
Currently when clicking the label of a textarea that has an id it will not focus the associated input. 
This pull request fixes the issue by using the same logic as the Input.php component.


robsontenorio:
Thanks!

PR #543, opened 2024-07-28
Colorpicker: add focus outline for input with type color

anfeichtinger:
Currently the color input of the Colorpicker has no focus styles. There is no feedback on whether it is focused or not.
This pull request adds the focus styles a primary input has.

image


robsontenorio:
Thanks!

PR #549, opened 2024-08-01
Resolve erroneous route matching in menus

hackerESQ:
Using the `activate-by-route` attribute will erroneously match links that have the same prefix. 

For example, these two links will show as active:

1. /users/1
2. /users/101

This is because of this line: 

    return $this->link != '/' && Str::startsWith($route, $link);

Which instead should be:

    return $this->link != '/' && Str::is($route, $link);

To avoid matching links that start with another matching link.

Screenshot for reference:

image

### DIFF
--- a/src/View/Components/MenuItem.php
+++ b/src/View/Components/MenuItem.php
@@ -45,7 +45,7 @@ class MenuItem extends Component
             return true;
         }
 
-        return $this->link != '/' && Str::startsWith($route, $link);
+        return $this->link != '/' && Str::is($route, $link);
     }
 
     public function render(): View|Closure|string


robsontenorio:
It breaks some expected cases ... Mainly when you deep navigate through same base prefix route.


http://localhost:8018/docs/components/menu#automatic-active-state

image

robsontenorio:
I advice you to use less ambiguous names for routes, specially on menus, in order to avoid collision.

hackerESQ:
I don't think this is due to ambiguous names. For example, if I have a user with ID # 2. Based on the `startsWith` logic, these would match too:

* **/users/2**3
* **/users/2**3/edit
* **/users/2**3/create
* etc

That's not what we want, right? And I'm not seeing how "less ambiguous names" solves this.

robsontenorio:
The current logic on `main` branch works for cases I mentioned above.

```html
// On your sidebar menu you have  this unique base link "/users"
<x-menu link="/users" > 

// You can use both on sidebar because its ambiguous  ...
<x-menu link="/users" > 
<x-menu link="/users/12" > 

```

hackerESQ:
Agreed 100%. But I'm not referring to unique sub-pages. I'm referring to conflicts driven by ID numbers that might `startWith` the same number. For example user with ID # 1 and user with ID # 12 will conflict since we're only matching the first digit of # 12. 

More concrete example: if I am currently viewing a user with ID # 1, the menu link for 

```php
<x-menu link="/users/**1**2" > 
```

will be marked as active. In which case I will have 2 or more links marked as active per my screenshot above:

image


Both the ID # 1 and ID # 11 were marked as active links in the menu.

robsontenorio:
Try this, it **won't** work with **your branch**.

```html
// web.php

Volt::route('/users', 'docs.installation');
Volt::route('/users/{id}/edit', 'docs.sidebar');
Volt::route('/users/{id}', 'docs.layout');

```


```html
// It WON'T keep this menu selected , but IT SHOULD. This is the expected behavior on stable version.

<x-menu activate-by-route>
      <x-menu-item label="Users" link="/users" />
</x-menu>
```

hackerESQ:
Understood. Would you be open to adding an optional `exact` attribute on menu items? That attribute would require the link and route to match exactly in order to show as active.

That leaves the default functionality you're describing - but also avoids issues with numeric IDs unintentionally matching.

I'll update the PR.

robsontenorio:
There is an option for exact route match take a look on source code if it works for you. 

If not , I will accept the PR if doesn’t break the current usage.

hackerESQ:
Submitted a clean PR that adds the new option. We can leave this one closed.

PR #550, opened 2024-08-01
All Fields: allow to customize hint CSS classes

flatcapco:
Ability to override the default hint classes.

### DIFF
--- a/src/View/Components/Checkbox.php
+++ b/src/View/Components/Checkbox.php
@@ -70,7 +70,7 @@ class Checkbox extends Component
 
                     <!-- HINT -->
                     @if($hint)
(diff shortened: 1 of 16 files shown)


robsontenorio:
You missed an important part. This instructs Tailwind to compile these default classes. It works for you because there are some local cache for these CSS classes.

Could please fix this?


```
x-classes="...."
```

See how we did it for error classes.

https://github.com/robsontenorio/mary/blob/a62c4d106f02501e4f561d6891e8ed838421c0e0/src/View/Components/Input.php#L190

flatcapco:
Really sorry about that @robsontenorio I've pushed that now

Issue #555, opened 2024-08-06
Drawer use of x-trap directive renders the rest of the page unfocusable


TobiasLounsbury:
### maryUI version

1.34.2

### daisyUI version

4.5.0

### Livewire version

3.4.6

### What browsers are affected?

Firefox, Chrome, Safari, Microsoft Edge

### What happened?

Use of the x-trap alpine directive makes it impossible to use the drawer along with other elements on the page at the same time. In the normal course of a menu that navigates away this is fine, but in my specific use case, I need to be able to trigger a modal that lives elsewhere on the page and the trap directive causes the modal to be un-focusable. 
My suggested fix is to make an optional flag to selectively not apply the trap directive. Defaulting to the same behavior as currently would not affect future upgrades but should make it simple enough to remove the directive when needed for projects that need to.

TobiasLounsbury:
If this is accepted I can also do a PR against the website repo to update the documentation as well.

PR #556, opened 2024-08-06
Drawer: make x-trap directive on the drawer optional

TobiasLounsbury:
Closes https://github.com/robsontenorio/mary/issues/555


robsontenorio:
Thanks!

image

PR #558, opened 2024-08-07
MenuItem: Adds optional `exact`  attribute on menu items

hackerESQ:
The `exact` attribute requires the link and route to match exactly in order to show as active in a menu.

This leaves the default functionality of fuzzy-ish matching parent routes, but also avoids issues with numeric IDs unintentionally matching (e.g. ID # 1 and ID # 11 should not match).

Example usage:

```
<x-menu-item title="{{ __('Transactions') }}" icon="o-banknotes" link="####" exact />
//                                                                             ^
```

Ref: #549


robsontenorio:
Thanks!

PR #576, opened 2024-08-16
Remove # default src in crop modal

### LINKED ISSUE #575: File component requests page in IMG placeholder
### maryUI version

1.35.3

### daisyUI version

4.12.10

### Livewire version

3.5.4

Chrome

### What happened?

When I have a File component with a preview like this:
```
<x-file wire:model="logo" accept="image/*">
    <img src="{{ route('businesses.logo', ['business' => $business]) }}" alt="Logo" class="w-20 h-20 object-cover rounded-lg"/>
</x-file>
```

The route in this case resolves to the correct thing, it's not empty and the page displays the preview as expected.
But I noticed that the page seems to take a long time to settle, like something else was being requested.

In the network tab (filtered to IMG) there was a strange request with a `text/html` mime type. odd.

Screenshot 2024-08-16 at 09 10 44

I whittled it down to this line in the file component. With `#` as the default `src` it requests the current page's HTML and tries to put that in the `IMG` tag.

https://github.com/robsontenorio/mary/blob/main/src/View/Components/File.php#L199

Screenshot 2024-08-16 at 09 08 22

Removing `#` resolves the problem, and the page loads much quicker. PR incoming.

datashaman:
Resolves #575 

Using `#` as `src` in an `img` element downloads the page.
It is a good placeholder for `a:href`, not `img:src`.

### DIFF
--- a/src/View/Components/File.php
+++ b/src/View/Components/File.php
@@ -196,7 +196,7 @@ class File extends Component
                         <!-- CROP MODAL -->
                         <div @click.prevent="" x-ref="crop" wire:ignore>
                             <x-mary-modal id="maryCrop{{ $uuid }}" x-ref="maryCrop" :title="$cropTitleText" separator class="backdrop-blur-sm" persistent @keydown.window.esc.prevent="">
-                                <img src="#" />
+                                <img src="" />
                                 <x-slot:actions>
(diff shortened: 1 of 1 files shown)


robsontenorio:
Thanks!

PR #594, opened 2024-08-26
Table: Added an option to disable link in a given column via 'disable-link' header

Kaz-:
I added an option in table headers to disable the global link from `<x-table link="/my-link" >` for a given column, useful when you want a column to have a custom link via `@scope('cell_.....')` 

To use it you just add the `'disableLink' => true` attribute to your header like this :

```php
 public function headers(): array
    {
        return [
            ['key' => 'id', 'label' => '#', 'class' => 'hidden lg:table-cell'],
            ['key' => 'name', 'label' => 'Permission', 'sortBy' => 'name', 'disableLink' => true] , // <--- here
            ['key' => 'description', 'label' => 'Description', 'sortBy' => 'description'],
        ];
    }
```

### DIFF
--- a/src/View/Components/Table.php
+++ b/src/View/Components/Table.php
@@ -86,6 +86,12 @@ class Table extends Component
         return $header['hidden'] ?? false;
     }
 
+    // Check if link should be shown in cell
+    public function hasLink(mixed $header): bool
+    {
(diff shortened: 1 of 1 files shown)


robsontenorio:
Thanks!

PR #596, opened 2024-08-28
MenuItems: handle no title 

c-perez:
When a <x-menu-item/> is created without a tittle, an empty div is rendered.
 
Screenshot from 2024-08-27 19-33-54
image

Fix
image


robsontenorio:
Thanks!

PR #601, opened 2024-09-01
Table: additional condition to display `perPage` selector

c-perez:
Hide perPage select on empty tables

Before
image

After
image


robsontenorio:
Thanks!

Issue #604, opened 2024-09-05
Dropdown component overflow issue in table


flatcapco:
### maryUI version

latest

### daisyUI version

latest

### Livewire version

latest

### What browsers are affected?

Chrome

### What happened?

When using the x-dropdown component in the action column of a x-table if there are more than 2 rows the top row is lost behind the overflow of the table.

https://capture.dropbox.com/nSQvwRIc0fN5TTZz

The table has a div with overflow-x-auto - if it is set to visible or unset then it fixes the issue. 

https://capture.dropbox.com/2u2eUYKa95sk1TlI

This may only be an issue if there is a table at the bottom of the page with an action due to the space alpine has to place the menu.

Hope that helps

robsontenorio:
Currently there is a limitation with CSS itself. Please, check this https://github.com/robsontenorio/mary/issues/483#issuecomment-2170852643

@c-perez  Is there something we can do in here?

DarkGhostHunter:
Can this instead be used through the [Popover API](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API)?

robsontenorio:
@DarkGhostHunter 

The Dropdown component uses AlpineJS anchor plugin.

But, the issue in here is about CSS limitation when working with `overflow-auto-x`, as I pointed on my latest comment above. We will sove this for now with PR #606

PR #606, opened 2024-09-06
Table: add property to set classes to table div container

c-perez:
Fix for #604


robsontenorio:
Thanks @c-perez 

Just added a trick to force Tailwind compile that class. Because Tailwind won't scan certain php blocks to compile it, ex: constructor parameters or PHP methods.

PR #607, opened 2024-09-06
Table: allow striped + expandable

djoks:
- Fixed issue with zebra striping not being applied to `table` when rows are `expandable`.
- Updated `nth-child` selectors to correctly handle every `4th` and `3rd` rows.
- Resolved conflicts between hover states and expanded rows using dynamic `::class` and `class` bindings.

## Bug Report:

MaryUI version: `^1.20`
Livewire version: `^3.5`

## Affected Browsers:

- Chrome
- Firefox
- Safari
- Edge

## What Happened?

- Zebra Striping Issue: Rows with the expandable feature were affected by zebra striping, causing inconsistent row styling, particularly on hover.

## Screencast:

https://github.com/user-attachments/assets/e69466ef-bd36-46ce-b7c2-0f1330765cd7

### DIFF
--- a/src/View/Components/Table.php
+++ b/src/View/Components/Table.php
@@ -228,8 +228,11 @@ class Table extends Component
                                 ->except('wire:model')
                                 ->class([
                                     'table',
-                                    'table-zebra' => $striped,
-                                    'cursor-pointer' => $attributes->hasAny(['@row-click', 'link'])
+                                    'table-zebra' => $striped && !$expandable,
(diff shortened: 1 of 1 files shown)


djoks:
@robsontenorio I implemented a fix for this using arbitrary values and variants in Tailwindcss. Please check it out.

robsontenorio:
Thanks!

I simplified the implementation to keep the same behavior expected for expanded style and use daisyUI utility classes instead of raw CSS variables.

Issue #512, opened 2024-06-29
support wire.modal.live for tables


cooffeeRequired:
✅ For sure you can paste screenshots.
        
❌ Please, don't paste code as screenshot. Paste as formatted markdown code.        

PHP      
        
```php
public bool $modal = false;
```
        
HTML/Blade (Table.php:200)
        
```blade
                    <table
                        {{
                            $attributes
                                ->except(['wire:model.live', 'wire:model'])
                                ->class([
                                    'table',
                                    'table-zebra' => $striped,
                                    'cursor-pointer' => $attributes->hasAny(['@row-click', 'link'])
                                ])
                        }}
                    >
```

```diff
* from
- ->except('wire:model')
* to
+ ->except(['wire:model.live', 'wire:model'])
```

robsontenorio:
What do you intend with this ?

cooffeeRequired:
Support live updated of selected.

PR #608, opened 2024-09-07
Table: fix wire model modifiers like wire:model.live

anfeichtinger:
- Fixes Issue #512 and other wire:model modifiers.
- Fixes "Undefined method 'pluck'" on ArrayAccess by converting to Collection.
- Fixes "Nested ternary expressions (without parentheses) deprecated in PHP 7.4. Targeting PHP 8.3.0." by wrapping in parentheses.


robsontenorio:
Thanks!

- Changed a bit for better handling for `getAllIds()`
- Also better handle for correct handling of expandable keys.

PR #614, opened 2024-09-12
Table: add `expandable-condition`

### LINKED ISSUE #597: For the table component consider making an expandableCondition.
For this example its based on if the follow_up_date is present in the model. Here is my table component

```
<x-table :headers="$headers"
             :rows="$prospects"
             :sort-by="$sortBy"
             expandable
             expandableCondition="follow_up_date"
             link="/prospects/{id}/edit"
             striped
             wire:key="prospect-table"
             wire:model="expanded"
             with-pagination>
```

image

i just added a new public property in the constructer
```
public mixed $expandableCondition = null
```
And used it in the table loop
```
  @if($expandable)
    @if($row[$expandableCondition])
      <td class="w-1 pe-0">
        <x-mary-icon
          name="o-chevron-down"
          ::class="isExpanded({{ data_get($row, $expandableKey) }}) || '-rotate-90 !text-current !bg-base-200'"
          class="cursor-pointer p-2 w-8 h-8 bg-base-300 rounded-lg"
          @click="toggleExpand({{ data_get($row, $expandableKey) }});" />
       </td>
    @endif
@endif
```

marcos-gonzales:
add expandable condition boolean to table class

<hr />

Closes #597

image

### DIFF
--- a/src/View/Components/Table.php
+++ b/src/View/Components/Table.php
@@ -25,7 +25,7 @@ class Table extends Component
         public ?string $selectableKey = 'id',
         public ?bool $expandable = false,
         public ?string $expandableKey = 'id',
-        public mixed $expandableCondition null,
+        public mixed $expandableCondition = null,
         public ?string $link = null,
         public ?bool $withPagination = false,
         public ?string $perPage = null,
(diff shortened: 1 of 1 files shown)


robsontenorio:
Thanks!

Issue #616, opened 2024-09-13
Blade directive @scope doesnt work when combined with Blade::stringable.


machacekmartin:
### maryUI version

1.35

### daisyUI version

-

### Livewire version

3.5

### What browsers are affected?

Firefox, Chrome, Safari, Microsoft Edge

### What happened?

Custom scope directive works as intended until we add our custom configuration of `Blade::stringable` into `AppServiceProvider`.

Steps to reproduce
- Install this package
- Use example carousel from docs with scope directive on any of your pages ( you can use any component that uses scope directive ).

```blade
@php
    $slides = [
        [
            'image' => '/photos/photo-1559703248-dcaaec9fab78.jpg',
            'title' => 'Front end developers',
        ],
        [
            'image' => '/photos/photo-1565098772267-60af42b81ef2.jpg',
            'title' => 'Full stack developers',
        ],
        [
            'image' => '/photos/photo-1494253109108-2e30c049369b.jpg',
            'title' => 'Hey!',
        ],
    ];
@endphp
 
<x-carousel :slides="$slides">
    @scope('content', $slide)
        <div class="mt-16 bg-red-500 font-bold text-white rounded p-2 w-fit mx-auto">
            {{ $slide['title'] }} - {{  $loop->index }}
        </div>
    @endscope
</x-carousel>
```

- No problems so far
- Register this example `Blade::stringable` into your `AppServiceProvider` using `boot` method
```php 
public function boot(): void
{
    Blade::stringable(function(\Illuminate\Support\Carbon $value) {
        return $value->format('d. m. Y');
    });
}
```
- Run `php artisan optimize:clear`
- Page now no longer works and an exception is thrown.

```
Internal Server Error

ErrorException
Undefined variable $__bladeCompiler
```

This is due to how scope directive is implemented, but I have not found any solution to this problem. Please let me know if you manage to fix it. I want to apply default formatting of dates in blade components, but it seems the custom scope directive is somehow conflicting with it. 

I also opened new issue in the source package of the scope directive, since it happens with his code as well.
https://github.com/konradkalemba/blade-components-scoped-slots
But i figured you may solve it sooner, since last release of his package was in 2020

PR #617, opened 2024-09-13
Fix missing $__bladeCompiler variable in scope directive

machacekmartin:
This fixes the Issue of scope directive throwing an exception with Blade::stringable configuration being present in AppServiceProvider.

https://github.com/robsontenorio/mary/issues/616


machacekmartin:
Author of the original scope package made a good point - when a directive that requires bladeCompiler is not used - it may throw an error (because the used $__bladeCompiler does not exist anywhere in the blade compiler), so it is recommended to create it by default.

https://github.com/konradkalemba/blade-components-scoped-slots/commit/d66b44a5858d52e2ff4b030dbdb4e23df78a4bf6#diff-dc4533e9cdd82fd75dfdc74283ec5d2bbe050827b0c3d65d9f5c82205f2bc9e9R39

PR #621, opened 2024-09-14
Toggle: add required symbol if given, add validation errors

### LINKED ISSUE #610: Toggle: Add support for required marker and validation messages
Currently the toggle component does not support the required marker or validation messages. As I often use this component in forms it would be really helpful to also support this. I can create a pull request if necessary. 

Here is an example of what this could look like from a different project:
image

anfeichtinger:
This will ...

- ...add the required symbol after the label text if the "required" attribute is given.
- ...add validation error messages configurable like on the checkbox component.

image

Edit: Closes #610


robsontenorio:
Thanks!

PR #622, opened 2024-09-14
Checkbox: add required symbol if given

anfeichtinger:
This will add the required symbol after the label if the "required" attribute is provided. This was missing and I noticed it while updating the Toggle component.

image


robsontenorio:
Thanks!

PR #627, opened 2024-09-19
Input: change nowdoc language to BLADE and replace comments with blade style

anfeichtinger:
As discussed on the official discord server this PR does the following in the Input component:

- Change the nowdoc language from `<<<'HTML'` to `<<<'BLADE'`.
- Change the comment style from `<!-- ... -->` to `{{-- ... --}}`.

Note: Because the Blade language in newdoc has a problem with alpinejs's `@` directives those should be changed to `x-on:` instead, otherwise this can break the syntax highlighting. This is the same thing as per [alpinejs docs](https://alpinejs.dev/directives/on#shorthand-syntax). Here is a comparison:

image
image

### DIFF
--- a/src/View/Components/Input.php
+++ b/src/View/Components/Input.php
@@ -142,7 +142,7 @@ class Input extends Component
 
                     {{-- CLEAR ICON  --}}
                     @if($clearable)
-                        <x-mary-icon y-on:click="$wire.set('{{ $modelName() }}', '', {{ json_encode($attributes->wire('model')->hasModifier('live')) }})"  name="o-x-mark" class="absolute top-1/2 end-3 -translate-y-1/2 cursor-pointer text-gray-400 hover:text-gray-600" />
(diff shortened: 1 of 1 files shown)


cooffeeRequired:
Thats right, and also the problems i there for ex.'

image

The hightling doesn't recongnize the $attributes..

robsontenorio:
Actually the syntax theme looks ok for me with “@“. It really depends on what theme people use.

anfeichtinger:
It really depends on what comes after the "@" symbol I think. On the input component it looks like this at the top:
image
That is the marker when you have unclosed brackets in my current theme.

And like this at the bottom, which is the same as the content of a string.
image

I guess this can be theme dependent but I think VSCode still gets confused by it either way.
If you don't want to use it like this it is also okay for me though.

robsontenorio:
Thanks @anfeichtinger !

anfeichtinger:
Let me know if this should be done for other components.

robsontenorio:
@anfeichtinger  You are killing with all PRs ❤️ . 

I really appreciate if you can give me a hand with bug fixes at first, after that I think you can work on other components.

PR #629, opened 2024-09-20
Introduce new component: Password

### LINKED ISSUE #552: Password component
I wanna create a password input with toggle input type feature. Currently, we can show the right icon, but we can't custom the action on click.

anfeichtinger:
This will add a new component called Password, closes #552.

I have made the component by copying the Input component, then removing all options that are not relevant and adding a button that can toggle the input type between 'password' and 'text'. Most of the settings from Input still work and the position & icons of the toggle can be customized.

These options have been removed: 

```php
public ?bool $money = false,
public ?string $locale = 'en-US',
```

The new options related to passwords are:

```php
public ?string $passwordIcon = 'o-lock-closed',
public ?string $passwordVisibleIcon = 'o-lock-open',
public ?bool $right = false,
public ?bool $onlyPassword = false,
```

Here are some examples of different configurations:
```blade
<x-password hint="Minimal example" wire:model="password1" />
<x-password label="Clearable inline" wire:model="password1" inline clearable />
<x-password label="Left icon right toggle" wire:model="password1" icon="o-face-smile" right />
<x-password label="Left icon right toggle clearable" wire:model="password1" icon="o-face-smile" right clearable />
<x-password label="Right icon left toggle" wire:model="password1" iconRight="o-face-smile" />
<x-password label="Right icon left toggle clearable" wire:model="password1" iconRight="o-face-smile" clearable />
<x-password label="Without toggle" wire:model="password1" only-password />
<x-password label="Custom icons" wire:model="password1" password-icon="o-eye-slash" password-visible-icon="o-eye" />
<x-password label="Prefix & suffix" wire:model="password1" prefix="Password" suffix="#" />
<x-password label="Validation like normal input" wire:model="password2" placeholder="Placeholders are supported" />
```

The examples will render like so:

image

### DIFF
--- a/src/MaryServiceProvider.php
+++ b/src/MaryServiceProvider.php
@@ -49,6 +49,7 @@ use Mary\View\Components\MenuTitle;
 use Mary\View\Components\Modal;
 use Mary\View\Components\Nav;
 use Mary\View\Components\Pagination;
+use Mary\View\Components\Password;
 use Mary\View\Components\Pin;
 use Mary\View\Components\Popover;
 use Mary\View\Components\Progress;
@@ -155,6 +156,7 @@ class MaryServiceProvider extends ServiceProvider
         Blade::component($prefix . 'main', Main::class);
         Blade::component($prefix . 'nav', Nav::class);
         Blade::component($prefix . 'pagination', Pagination::class);
+        Blade::component($prefix . 'password', Password::class);
         Blade::component($prefix . 'pin', Pin::class);
         Blade::component($prefix . 'popover', Popover::class);
         Blade::component($prefix . 'progress', Progress::class);
--- a/src/View/Components/Password.php
+++ b/src/View/Components/Password.php
@@ -1,0 +1,214 @@
+<?php
+
+namespace Mary\View\Components;
+
+use Closure;
+use Exception;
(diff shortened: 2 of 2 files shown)


robsontenorio:
I think a good feature is to toggle password visibility by clicking at “eye” icon. So , this icon should be toggled when clicked.

anfeichtinger:
I'm not sure what you mean. Toggling between the password visibility by clicking on the icon is already implemented, though the default icon in this pull request is the lock icon. If that is the issue the default can easily be changed to the eye icon.

elijahcruz12:
> I'm not sure what you mean. Toggling between the password visibility by clicking on the icon is already implemented, though the default icon in this pull request is the lock icon. If that is the issue the default can easily be changed to the eye icon.

I think having an eye as the default would be a more proper choice in terms of visibility, as that is what most sites use.

robsontenorio:
@anfeichtinger sorry. It works 👍

anfeichtinger:
I will change the default icons to `o-eye` and `o-eye-slash` respectively today. Then this should be ready to be merged.

robsontenorio:
Thanks @anfeichtinger !

PR #636, opened 2024-09-21
MenuItem: add support for modifiers on wire:navigate

anfeichtinger:
As per a comment on discord this PR will add support for modifiers on `wire:navigate` like `wire:navigate.hover` to the MenuItem component.


robsontenorio:
Thanks @anfeichtinger !

PR #644, opened 2024-09-25
Calendar: allow duplicated dates

groxw:
Hello, 

I found a bug, If there's an event has a same date, it shows only the first event, I am trying to show all event and seperated by 
`<hr>`

I am using foreach, because it takes bit faster than using map, [ref](https://stackoverflow.com/questions/18144782/performance-of-foreach-array-map-with-lambda-and-array-map-with-static-function)


groxw:
Screenshot_49
It will show like that

robsontenorio:
Just made it simpler. Thanks!

PR #645, opened 2024-09-26
Choices: add support for placeholders

### LINKED ISSUE #640: Add Choices Placeholder
Can you add the possibility of putting a placeorder as for a seclect please?

anfeichtinger:
Adds support for providing `placeholder="Some value"` to Choices and ChoicesOffline. Closes #640.

I tried using the placeholder for the input first but that lead to layout issues so I've added an absolute positioned span instead. I tried to make it look as similar as possible to the input placeholder. 

Because this is a complex component it should be tested with different use-cases. I tested it by adding the placeholder to all cases on the mary-ui.com page for choices. The only not so perfect thing I found was that when searching with the Choices component, the placeholder is removed after the debounce, leading it to overlap with the inputted text for 250ms.

image

### DIFF
--- a/src/View/Components/Choices.php
+++ b/src/View/Components/Choices.php
@@ -101,8 +101,7 @@ class Choices extends Component
                             isDisabled: {{ json_encode($isDisabled()) }},
                             isRequired: {{ json_encode($isRequired()) }},
                             minChars: {{ $minChars }},
-                            searching: false,
-
+                            
                             init() {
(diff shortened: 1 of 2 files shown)


robsontenorio:
Thanks @anfeichtinger 

Just made things simpler and used the `focused` state to control the visibility.
Also replaced CSS classes for RTL support (start/end) .

PR #647, opened 2024-09-27
Choices/ChoicesOffline: add keyboard support

akhmads:
- Press enter key to trigger dropdown
- Support tab, top/bottom arrow key button to select choice item
- Press enter key to select choice item
- Prevent arrow key (top, left, right, down) and tab from triggering search process

https://github.com/user-attachments/assets/b85a36ed-359f-434d-ba34-cd8f5f6269a1

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
     "require": {
         "illuminate/support": "^10.0|^11.0",
         "jfcherng/php-diff": "^6.15",
-        "laravel/prompts": "^0.1.15",
+        "laravel/prompts": "^0|^1",
         "blade-ui-kit/blade-heroicons": "^2.3"
     },
     "require-dev": {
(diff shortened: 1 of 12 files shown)


quileab:
Good work, waiting for PR merge

robsontenorio:
@akhmads 

- It would more intuitive to make it work with "arrow down", as a first trigger. 
- The "enter" tigger is problem, because it will submit the form depending on where it is placed. So, "arrow down" it would be fine.

Could you please handle it?

When I click on this input it shows me the list by default (it is ok), but "arrow down" does not work. Just "tab", then "arrow down" works.

image

akhmads:
Ok sir, I will try to fix it.

akhmads:
I have made "down arrow" a dropdown trigger, please review
Thank you

robsontenorio:
@akhmads I came up with another approach. Please, give a try.

PR #648, opened 2024-09-27
MenuItem: adds icon right and spinner

weelrs8:
`This is my first time contributing to a project, forgive me if I make a mistake, point out the error so I can improve in the future.`

I added support for the **icon on the right**.

I added support for the **spinner** .

When using a Dropdown, for example, and the action takes a few seconds to complete, it is crucial that the user has visual feedback that something is happening.
To obtain the expected result it is necessary to use MenuItem with `wire:click.stop="archive"`

Captura de tela de 2024-09-27 11-45-32 Captura de tela de 2024-09-27 11-45-23

### DIFF
--- a/src/View/Components/MenuItem.php
+++ b/src/View/Components/MenuItem.php
@@ -4,7 +4,6 @@ namespace Mary\View\Components;
 
 use Closure;
 use Illuminate\Contracts\View\View;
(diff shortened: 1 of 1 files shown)


robsontenorio:
Thanks! 

Just removed the "right" icon because it didn't look good with many menus with different size of texts.

PR #651, opened 2024-09-28
Revert "MenuItem: add support for modifiers on wire:navigate"

robsontenorio:
Reverts robsontenorio/Mary#636

@anfeichtinger Unfortunately I need revert this, because I noticed it breaks the default behavior of menu items that is `wire:navigate`

### DIFF
--- a/src/View/Components/MenuItem.php
+++ b/src/View/Components/MenuItem.php
@@ -75,7 +75,7 @@ class MenuItem extends Component
                             @endif
 
                             @if(!$external && !$noWireNavigate)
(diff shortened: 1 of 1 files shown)


anfeichtinger:
That's strange, I checked the html and it was the same. Sorry for the trouble!

anfeichtinger:
@robsontenorio Could this be fixed by using `{{ $attributes->wire('navigate') ?? 'wire:navigate' }}` instead. I was not aware it should be used by default. I can make a new pull request if that is ok.

robsontenorio:
@anfeichtinger you can give a try. 👍

Issue #649, opened 2024-09-27
Ranged Flatpickr in Drawer Issue


flatcapco:
### maryUI version

1.35.9

### daisyUI version

4.12.10

### Livewire version

3.5.8

### What browsers are affected?

Chrome

### What happened?

There appears to be an issue with a ranged flatpickr .

If you click the 2nd date the flatpickr closes and you have to reopen it to then click the second date successfully.

This is in chrome:
https://capture.dropbox.com/6Km2Sf0GbRzBix49

```php

                <x-datepicker
                :label="__('Date Range')"
                wire:model.live.debounce="dates"
                :config="['mode' => 'range', 'altFormat' => 'd-m-Y']"
                inline
            />
```

For clarity - the issue also happens when outside of a drawer

itsmenewbie03:
The cause of the issue is the `.live` modifier. Removing it makes the `datepicker` work as expected.

flatcapco:
@itsmenewbie03 I'm actually still experiencing the issue with just wire:model - are you able to confirm that too?

PR #653, opened 2024-09-29
feat: live mode for datepicker range mode

itsmenewbie03:
Using `wire:model.live` makes datepicker with mode set to `range` behave unexpectedly as described in #649. This is due to livewire sending update as long as the user selects the first date. I can't find anyway to customize the behavior of livewire's `.live` modefier. My proposed fix for instances when you want live update as long as the user selects a range is to add a `live` attribute to datepicker. This will add a change event listener using `x-on:change` which checks if the date is complete by spliting it with `'to'` and checking if the resulting array contains two items. The listener only gets attached if the mode is set to range, hence making the `live` attribute range mode exclusive.

### DIFF
--- a/src/View/Components/DatePicker.php
+++ b/src/View/Components/DatePicker.php
@@ -18,6 +18,7 @@ class DatePicker extends Component
         public ?string $hint = null,
         public ?string $hintClass = 'label-text-alt text-gray-400 py-1 pb-0',
         public ?bool $inline = false,
+        public ?bool $live = false,
         public ?array $config = [],
         // Validations
         public ?string $errorField = null,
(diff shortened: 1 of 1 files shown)


robsontenorio:
Hey @itsmenewbie03 , appreciate that!

But to keep consistence with Livewire syntax, like all other components, we need to find some workaround to make it work with “wire:model.live”

itsmenewbie03:
> Hey @itsmenewbie03 , appreciate that!
> 
> But to keep consistence with Livewire syntax, like all other components, we need to find some workaround to make it work with “wire:model.live”

Thanks, that's what I strived for at first but turns out I can't xD. I checked through the Livewire docs  and found nothing helpful for the case. Flatpikr provides some event hooks but I can't figure out how to cancel the change during the first date selection. I'd love to see what's gonna be the workaround to have the desired behavior when using `wire:model.live`.

PR #662, opened 2024-10-03
Swap: introduce as new component

anfeichtinger:
This will add a new component Swap, built to simplify [daisy-ui's swap functionality](https://daisyui.com/components/swap/).
Currently we only have the theme toggle which is probably the most common use case for swapping, but it is limited to swapping the current theme. With this you can swap any content. Here are some examples:

```blade
{{-- Default works without model and uses theme toggle icons --}}
<x-swap />

{{-- wire:model + custom icons + flip animation + custom size --}}
<x-swap wire:model="rating" true-icon="o-speaker-wave" false-icon="o-speaker-x-mark"
             icon-size="h-8 w-8" class="swap-flip" />

{{-- wire:model.live + simple text content (size and color can be customized by using a parent element) --}}
<x-swap wire:model.live="rating" true="ON" false="OFF" />

{{-- Completely custom content --}}
<x-swap>
    
    {{-- True--}}
    <x-slot:true
        class="h-12 p-2 bg-primary text-primary-content rounded-2xl flex flex-row items-center justify-center">
            Custom on
    </x-slot:true>

    {{-- False--}}
    <x-slot:false
        class="h-12 p-2 bg-error text-error-content rounded flex flex-row items-center justify-center">
            Custom off
    </x-slot:false>

</x-swap>

{{-- Before and after slots --}}
<x-swap wire:model.live="rating" class="flex flex-row space-x-2 w-fit">

    {{-- Before --}}
    <x-slot:before class="text-primary">
        Swap me!
    </x-slot:before>

    {{-- After --}}
    <x-slot:after class="text-error">
        {{ $this->rating ? 'ON' : 'OFF' }}
    </x-slot:after>

</x-swap>
```

It is kind of hard to showcase as there are animations involved but here are the above examples in their true and false states as screenshots:

True             |  False
:-------------------------:|:-------------------------:
image  |  image

Note: The indeterminate state is currently not supported as it can only be set via javascript. If it is implemented we could switch between 3 states.

### DIFF
--- a/src/MaryServiceProvider.php
+++ b/src/MaryServiceProvider.php
@@ -64,6 +64,7 @@ use Mary\View\Components\Spotlight;
 use Mary\View\Components\Stat;
 use Mary\View\Components\Step;
 use Mary\View\Components\Steps;
+use Mary\View\Components\Swap;
 use Mary\View\Components\Tab;
 use Mary\View\Components\Table;
 use Mary\View\Components\Tabs;
@@ -170,6 +171,7 @@ class MaryServiceProvider extends ServiceProvider
         Blade::component($prefix . 'stat', Stat::class);
         Blade::component($prefix . 'steps', Steps::class);
         Blade::component($prefix . 'step', Step::class);
+        Blade::component($prefix . 'swap', Swap::class);
         Blade::component($prefix . 'table', Table::class);
         Blade::component($prefix . 'tab', Tab::class);
         Blade::component($prefix . 'tabs', Tabs::class);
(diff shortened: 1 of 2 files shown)


robsontenorio:
Great!

I don’t think that indeterminate state would be necessary. Because, it would not make sense in here. 

If you have extra time , please, consider adding some docs for this. On mary docs repo. It can be placed at “UI” section.

robsontenorio:
@anfeichtinger Thanks!

PR #665, opened 2024-10-06
Improve Table Selectable

c-perez:
When a table is empty, the SelectAll checkbox is always active.

This commit disables and uncheck SelectAll checkbox for a better UI experience.

Before
image

After
image


robsontenorio:
Thanks! @c-perez

PR #667, opened 2024-10-10
Add close button to Modal component

iurigustavo:
Introduce a close button inside the modal box. The button supports both form submissions and live wire updates, depending on the presence of an ID, enhancing the modal's usability and user experience.

image


robsontenorio:
Thanks! @iurigustavo

PR #677, opened 2024-10-22
Extend the MenuItem class to support multiple badges, including icons.

dfrl-dev:
Added the ability to add an array of badges, including support for the heroicons.

image

Usage:  

- Pass an array of arrays through the badges parameter. Possible keys are 'label', 'icon', 'classes'.
- If you pass an array, it overrides the single badge option, but for backwards compatibility the single badge approach is still supported.

```
<x-menu-item 
            :title="MenuItem Title" 
            :badges="[
                ['label' => "Badge1 Label", 'classes' => "!badge-success"],
                ['icon' => 'o-cloud', 'classes' => '!badge-error'],
            ]" 
            icon="o-document-text"
            />
```

### DIFF
--- a/src/View/Components/MenuItem.php
+++ b/src/View/Components/MenuItem.php
@@ -4,7 +4,6 @@ namespace Mary\View\Components;
 
 use Closure;
 use Illuminate\Contracts\View\View;
-use Illuminate\Support\Facades\Route;
 use Illuminate\Support\Str;
 use Illuminate\View\Component;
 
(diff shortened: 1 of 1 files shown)


dfrl-dev:
That's fair.  

So in other words, use the menu item without a specific title, and create a slot that has spans for each state/badge I'd like to display?

I have to ask: why disable the slot from being rendered if there's a title present?

dfrl-dev:
Ahh actually -- this may be a seperate issue entirely, but the $slot variable used in the render function isn't even defined, so slots weren't working on menu items at all.

dfrl-dev:
Ah I see. 

I'll implement that based on how you have it in ListItem -- it'd be very useful to me in my application, and I'd love to contribute.

PR #678, opened 2024-10-22
Added member variable for  so that it renders when used with MenuItem

dfrl-dev:
Simply added the $slot variable to MenuItem.php so that the $slot can actually be used.

Carrying on the way the variable is named in the class already, usage looks like this:

```
<x-menu-item>
     <x-slot:slot> Slot Content </x-slot:slot>
</x-menu-item>
```


robsontenorio:
@dfrl-dev  After checking the current component it just supports the default content slot. I will add this example to docs.

image

PR #685, opened 2024-10-27
feat: make input inside a modal autofucosable

### LINKED ISSUE #683: autofocus
“It would be helpful to have an option to enable autofocus on an input field. When opening a modal with a single input field, setting it to autofocus would allow users to start typing immediately without needing to click on the field first.”

itsmenewbie03:
This commit makes an input inside a modal autofocusable by adding the `focus_event` attribute to the modal and `modal_autofocus` attribute to the input which should be focused. This has to be done as the autofocus HTML attribute only sets focus for inputs that are visible on page load which are not the case for most modal.

closes #683

### DIFF
--- a/src/View/Components/Input.php
+++ b/src/View/Components/Input.php
@@ -21,6 +21,7 @@ class Input extends Component
         public ?bool $inline = false,
         public ?bool $clearable = false,
         public ?bool $money = false,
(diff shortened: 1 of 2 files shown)


itsmenewbie03:
will be closed in favor of #686

PR #696, opened 2024-11-06
Progress: support for decimals

roberto-proj:
When there is a decimal value such as "10.5", the progress bar will be able to assign its decimal fraction to it, instead of considering only the integral part of that number.


robsontenorio:
Thanks!

PR #699, opened 2024-11-08
Main: use `persist` approach for sidebar state

cooffeeRequired:
removing redundant code, there is no need to call fetch and run the whole event via $wire, now it is no longer needed, $persists from Alpine JS was used and the whole event works better

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -4,12 +4,6 @@ use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\Facades\Storage;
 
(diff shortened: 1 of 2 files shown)


robsontenorio:
I have tried this approach with `persist` in the past, but it does work well.
When you first open the app or refresh it , the side bar flashes.

This happens because the screen is rendered before JS can be evaluated.
 So I ended up with session/route approach, so things happens on server side.

https://github.com/user-attachments/assets/bb47aabb-2ea3-4892-9bb9-1255cff4487d

cooffeeRequired:
did you try x-cloak?

PR #707, opened 2024-11-13
Add column visibility control to Table component

iurigustavo:
Introduced TableColumnsSelector component to manage column visibility settings using localStorage. Updated the Table component to respect these settings and dynamically show/hide columns based on user preferences.


robsontenorio:
@iurigustavo when you have chance update this with the main branch to check if is ok.

iurigustavo:
It still needs a few more tests, but I'll write it down here.

PR #711, opened 2024-11-15
FEATURE | Radio - New label option to integrate HTML

9christian9:
I added the possibility of passing HTML in the option label.

```PHP
<x-code>
        @verbatim('docs')
            @php
                $users = App\Models\User::take(3)->get();
                $labelHtml = "<x-popover>
                    <x-slot:trigger>
                        <x-icon name='o-information-circle'/> Select one
                    </x-slot:trigger>
                    <x-slot:content>
                        HEY, this is a suggestion
                    </x-slot:content>
                </x-popover>";
            @endphp
            <x-radio
                :labelHtml=$labelHtml
                :options="$users"
                wire:model="selectedUser3" />
        @endverbatim
    </x-code>
```

### DIFF
--- a/src/View/Components/Radio.php
+++ b/src/View/Components/Radio.php
@@ -13,6 +13,7 @@ class Radio extends Component
 
     public function __construct(
         public ?string $label = null,
+        public ?string $labelHtml = null,
         public ?string $hint = null,
         public ?string $hintClass = 'label-text-alt text-gray-400 ps-1 mt-2',
         public ?string $optionValue = 'id',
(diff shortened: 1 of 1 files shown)


9christian9:
mary-ui.com | Related DOC https://github.com/robsontenorio/mary-ui.com/pull/51

9christian9:
@robsontenorio what do you think? Will it be implemented?

robsontenorio:
Thanks for this!

robsontenorio:
@9christian9  The pitfall of this approach is that it will apply that HTML to all options. So, the popover content (or any html put) will be the same. A little gain in here.

With the same amount of code you can just use the label `slot` to achieve the same.

```
```


image

Or even simpler, just use the hint.


image

9christian9:
Ciao @robsontenorio, The purpose of this PR is precisely to pass HTML from a variable to the component because in my personal project I need this.
In fact, in my PR, the $labelHtml variable is quite distinct from the normal $label variable.
I think this modification may also be useful for other people

robsontenorio:
Thanks for your contribution!

With the release of maryUI v2, I’m closing this. If you still think this should be merged, please create a new branch from main and submit it again.

Issue #672, opened 2024-10-17
add support to change heading tags on cards


elliottcoe:
It would be great if the headings on a card could be changed to a html heading (h1, h2, h3 etc), this would really help with SEO when the card is the main part of the content on page.

9christian9:
Hi, I opened this PR https://github.com/robsontenorio/mary/pull/712

robsontenorio:
Closing in favor of #736

nickwild-999:
Shame you are not considering this @robsontenorio an h tag make much more sematic sense to be able to implement rather than an alt tag  #736 they serve different purposes. Can you be persuaded? 😊

PR #623, opened 2024-09-17
maryUI v2

robsontenorio:
# Overview

We don't expect breaking changes, except for removal of the default "primary" style for all the things.

- [x] Initial migration
- [x] Upgrade guide
- [x] Docs site
- [x] Review components
- [x] Review the Bootcamp
- [ ] Update all demos
- [x] Migrate the "Choices" component.
- [x] Review installer
- [x] Launch beta1 - **March 14**
- [x] Another betas - **March 15+**
- [x] Release v2 - **31 march**
- [ ] Release a brand new cool demo 

# Default style

Currently all the things are styled as "primary". That was a mistake and it will be changed at 2.x. According to UX/UI experts, by default, we should not use the primary style for all the things.

But if you want to keep all the things as primary , just set it as primary.

```html
<x-input ... class="input-primary" />
<x-toggle ... class="toggle-primary" />
<x-checkbox ... class="checkbox-primary" />
```

# Changelog

- Make all things non-primary **(breaking)**
- Uses default daisyUI classes instead of `text-red-*` , `text-gray-*`
- WIP ...

PR #712, opened 2024-11-15
FEATURE | Choose the html tag <h> to improve SEO

9christian9:
This feature allows you to decide which <code>h</code> HTML tags to use to improve SEO

### DIFF
--- a/src/View/Components/Card.php
+++ b/src/View/Components/Card.php
@@ -12,7 +12,9 @@ class Card extends Component
 
     public function __construct(
         public ?string $title = null,
+        public ?string $titleSEO = null,
         public ?string $subtitle = null,
+        public ?string $subtitleSEO = null,
         public ?bool $separator = false,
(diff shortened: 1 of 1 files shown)


9christian9:
#672

9christian9:
@robsontenorio what do you think? Will it be implemented?

igergow (changes requested):
That's good idea, I was going to suggest something similar, can you make it more precise titleSEO can be something as heading/headingType and the subtitleSEO as subheading/subheadingType?

robsontenorio:
Thanks for this!

elliottcoe:
> Thanks for this!
> 
> I am delaying it for now because of upcoming maryUI v2 (daisyUI 5 + Tailwind 4).

Oh that's great! Wasn't aware of MaryUI V2, exciting 😄

robsontenorio:
See #623

PR #718, opened 2024-11-19
Steps: add vertical steps support

Dhalion:
Adds dynamic class to support vertical steps


robsontenorio:
Thanks for this!

robsontenorio:
Yes, this works. But unfortunately there is a problem about layout consistency: there are many places to put the "step content", that we can not implement  it automatically . In contrast of  horizontal  approach the content goes bellows the steps.

robsontenorio:
Ok, now I got it. The best solution in here is to use the Timeline component instead.

PR #726, opened 2024-11-24
Input: Easier label styling

v-stamenova:
Allows for easier customisation of the label (both inline and standard)


robsontenorio:
Thanks for this!

robsontenorio:
@v-stamenova Unfortunately this will cause a issue, because you put two `@class()` on same component. That means this ...

```
<x-input class="bg-red-500" />
```

... will apply `bg-red-500` at all `@class()` inside the component.

PR #733, opened 2024-12-05
Fix for pagination "onEachSide"

romanwuensche:
`$rows->onEachSide(1)` was not working as expected in my project. Seems we have to use the full qualified namespace in the template for this to work:

before
after

### DIFF
--- a/src/View/Components/Pagination.php
+++ b/src/View/Components/Pagination.php
@@ -46,7 +46,7 @@ class Pagination extends Component
                     </div>
                     @endif
                     <div class="w-full">
-                    @if($rows instanceof LengthAwarePaginator)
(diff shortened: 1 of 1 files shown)


robsontenorio:
Thanks for this!

robsontenorio:
@romanwuensche I can't reproduce that using this simple query 

```
$users = User::paginate(3);
```

How does it behaves for you?

romanwuensche:
Hi @robsontenorio. Still the same behavoir on my end. I will try a fresh install without anything special.

robsontenorio:
Thanks for your contribution!

With the release of maryUI v2, I’m closing this. If you still think this should be merged, please create a new branch from main and submit it again.

Issue #745, opened 2024-12-27
Option list flashes


TobiasLounsbury:
### maryUI version

1.41.1

### daisyUI version

4.5.0

### Livewire version

3.5.12

### What browsers are affected?

Firefox, Chrome, Safari, Microsoft Edge

### What happened?

On a page with a choices widget, on initial page load, for a brief moment while alpine is being bootstrapped the options list is visible. This only happens on choices, not on choices-offline as that options list has an x-cloak directive added. 

PR to add x-cloak incoming.

PR #746, opened 2024-12-27
Added x-cloak to choices

TobiasLounsbury:
Added an x-cloak directive to the choices option list container so it doesn't flash momentarily on page load.

https://github.com/robsontenorio/mary/issues/745


robsontenorio:
Thanks Tobias!

PR #756, opened 2025-01-10
Table: add row-selection-all event to table

tashuo:
Add toggleCheckAll event to the table component


robsontenorio:
@tashuo Unfortunately  It could not be simple as this, because it will cause false positive. 
In general "check all" behavior must match with checkbox style in two states : "current page" and "all pages".

Take a look on how Filament behavior works with "check all"

Issue #624, opened 2024-09-18
Very clear table header text (in light themes)


weelrs8:
My OS: Ubuntu 24.04.1 LTS

### maryUI version

1.35.9

### daisyUI version

4.12.10

### Livewire version

3.5.6

### What browsers are affected?

Firefox, Chrome

### What happened?

When using a light theme, the texts in the headers table are too light

Dark:
![Captura de tela de 2024-09-18 08-35-49](https://github.com/user-attachments/assets/791059f2-6809-4329-afd8-913c9dcea2e6)

Light:
![Captura de tela de 2024-09-18 08-35-57](https://github.com/user-attachments/assets/8384ef2b-ea5d-4c42-9982-fe810b7931fa)

To reproduce the problem, just go to https://mary-ui.com/docs/components/table and change the theme to light.

robsontenorio:
Maybe some browser plugin or tweak on your OS?

Here I can’t reproduce it. Anyone ?

anfeichtinger:
For me it is the same on the docs page.

![image](https://github.com/user-attachments/assets/d575cc3c-8ade-42a9-9c0a-398ae19c2ba8)

From left to right: Chrome (No plugins), Brave (Ad-Blocking), Firefox (Ad-Blocking)

Versions:

- Windows 11 Pro 23H2 Build 22631.4169
- Google Chrome 128.0.6613.138 (Official Build) (64-bit)
- Brave 1.69.168 Chromium: 128.0.6613.138 (Official Build) (64-bit)
- Firefox 128.0.3 (64-bit)

On my page for testing mary-ui I get full black on the header without setting any color.

```
@php
        $users = Illuminate\Support\Collection::make([
            0 => ['id' => 1, 'name' => 'Andreas Feichtinger', 'city' => ['name' => 'Kalsdorf bei Graz']],
            1 => ['id' => 2, 'name' => 'Jon Doe', 'city' => ['name' => 'Graz']],
            2 => ['id' => 3, 'name' => 'Jane Doe', 'city' => ['name' => 'Wien']],
        ]);

        $headers = [
            ['key' => 'id', 'label' => '#'],
            ['key' => 'name', 'label' => 'Nice Name'],
            ['key' => 'city.name', 'label' => 'City'],
        ];
    @endphp
    <div class="container-card col-span-full min-w-full">
        <x-header title="Table" class="col-span-full" />
        <div class="col-span-full -mt-10 space-y-6">
[...]

Issue #373, opened 2024-04-11
Table labels unreadable with dark themes


bestmomo:
### maryUI version

1.26

### daisyUI version

4.10.1

### Livewire version

3.4

### What browsers are affected?

Firefox

### What happened?

![Capture d'écran 2024-04-11 181805](https://github.com/robsontenorio/mary/assets/2959682/e013989e-5221-49ad-9678-df6a2dbbdcbd)

robsontenorio:
What is your OS ? This issue also happens on Chrome ?

willcastillo:
I can confirm it also happens in Chrome (Version 123.0.6312.106 (Official Build) (64-bit) / Win11)

bestmomo:
Also happens on Brave and Edge

robsontenorio:
Could you guys please check this?

- Bootstrap a brand new project at  http://phpsandbox.io

- Then, open a terminal on (+) icon

<img width="568" alt="image" src="https://github.com/robsontenorio/mary/assets/118955/a18ff670-aa5b-440c-8090-ceb9683c852f">

<br /><br />

- Install mary + yarn build

```
 composer require robsontenorio/mary && php  artisan mary:install && yarn build
```
<br />

- Then , refresh the page
<br />
<img width="607" alt="image" src="https://github.com/robsontenorio/mary/assets/118955/966dd61d-e2a6-4ffa-bdc8-1442b12a1877">

bestmomo:
Hello, I have an issue to access to [http://phpsandbox.io/](http://phpsandbox.io/) : PR_CONNECT_RESET_ERROR

bestmomo:
Still getting message PR_CONNECT_RESET_ERROR

willcastillo:
For what it's worth,

willcastillo:
For what it's worth, I see my table (specifically the THEAD elem) is missing this style that is present in the [first example in the mary-ui docs for table](https://mary-ui.com/docs/components/table#simple). Not sure what is causing that.

![mary-ui-table-missing-style](https://github.com/robsontenorio/mary/assets/1260357/9982db5b-fa3c-4808-9d2e-be0cf9f9642d)

bestmomo:
I solved it temporarily in src/Views/Components/Table.php on line 182 :
```
<!-- <thead @class(["text-black dark:text-gray-200", "hidden" => $noHeaders])> -->
<thead @class(["hidden" => $noHeaders])>
```

[...]

Issue #302, opened 2024-02-17
[Bug]: Table component headers not theme compatible


IamSAB:
### MaryUI version

1.18

### daisyUI version

4.6

### Livewire version

3.3

### What browsers are affected?

Firefox

### What happened?

![grafik](https://github.com/robsontenorio/mary/assets/8931240/dc32e9bf-72f7-4e43-bf34-9d367a025a2e)
![grafik](https://github.com/robsontenorio/mary/assets/8931240/ac0ddfcb-8d5a-4cff-b2ab-921781826a8a)

The table component uses a fixed text color for table headers `text-black dark:text-gray-200` .
However, daisyUI is using CSS variables for themes not `dark:`, what allows for multiple themes.
[Theming with DaisyUI](
https://github.com/saadeghi/daisyui/discussions/640)

Overall, I would propose to not use any Tailwind colors (which are not themed) in components (e.g. text-gray-200) but classes used in DaisyUI (e.g bg-neutral, text-primary, text-content etc). If you do that consisten, all components are themed.
Whoever does not used DaisyUI and only Tailwind could then also style them using @apply.
This is for most components the case, but for some not, mainly defining text colors.

What do you think?

robsontenorio:
@IamSAB 

Are you able to send a PR to this specific use case ?

KozSuper:
Using this can be applied to multiple dark themes.
```js
darkMode: ['variant', [
      '@media (prefers-color-scheme: dark) { &:not(.light *) }',
      '&:is(.dark *)',
    ]]
```

robsontenorio:
That is solved on latest versions ...

DE-dylan-lauzon:
This isn't fixed with the custom themes from DaisyUI, the bug is visible when using the aqua theme that is used by the custom theme changer in the docs:

https://github.com/user-attachments/assets/d586b9fc-c2b9-4073-bacf-abc6c1ebca2d

Only with the default light and dark themes does the table header respect the theme, from what I can tell.

PR #797, opened 2025-02-18
fixed table header text color

makroxyz:
fix for  #373 #624 #302


makroxyz:
In the commit I removed fixed color classes...

robsontenorio:
Thanks!

It will be definitely fixed on v2.

Available on v2 beta. 

https://v2.mary-ui.com

PR #800, opened 2025-02-27
Add Laravel 12 support

robsontenorio:


DenaSon:
Add support for laravel 12 faster.pls

PR #812, opened 2025-03-15
fix: rounded styling of toast component

ChrisToxz:
moved `rounded-md` to the correct `<div>`.
Now the toast is rounded as intended

### DIFF
--- a/src/View/Components/Toast.php
+++ b/src/View/Components/Toast.php
@@ -29,13 +29,13 @@ class Toast extends Component
                                     "
                 >
                     <div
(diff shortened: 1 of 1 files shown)


ChrisToxz:
I have been wrong here. Just learned that each theme of daisyUI are not just colors but also border width, rounded attributes and so on differs between themes.

A suggestion would be to be able to change theme in the docs as well, so you know exactly what to expect with the theme you are using.

robsontenorio:
Yes , theme stuff comes from daisy. And its docs has a theme generator and a bunch of variables available.

PR #817, opened 2025-03-17
"Fix select class usage to align with daisyUI's recommendation"

s-kitade:
"Fix to match the intended usage of daisyUI.
Instead of adding the select class to the label tag, apply the select class to the select tag."

## before
image

## after
image


robsontenorio:
This placement is intended for better composition with other elements. 

For now use daisy 5.0.3. 
There is an issue open on daisy.

robsontenorio:
Fixed right now by daisy 5.0.5

PR #822, opened 2025-03-20
Render $description for HTML

alloylab:
add support for html in $description


robsontenorio:
Hey,  this code is not safe because it uses {!! !!} (double exclamation marks) in Blade, which renders the content without escaping. If `$description` contains user-generated content, it could lead to Cross-Site Scripting (XSS) vulnerabilities.

alloylab:
@robsontenorio I agree its not safe for user-generated content, I wouldn't think a timeline would have user generated content... but I guess it could... I will close it out

PR #842, opened 2025-04-02
Add public `$id` property to all constructors where an `$uuid` property exists

### LINKED ISSUE #834: Allow custom id on all (form & input) components
Currently it is possible to add a custom ID e. g. for the [Checkbox component](https://github.com/robsontenorio/mary/blob/main/src/View/Components/Checkbox.php#L26). It would be great to also allow this on all other components, but at least on those that will be used in forms.

We need it for example to easily test, that a component is rendered conditionally. Also it can help reduce bugs if there are two instances of the same component with the same parameters. (Like in [this issue](https://github.com/robsontenorio/mary/issues/698), where it was added for checkboxes)

I can create a PR if needed.

anfeichtinger:
Like discussed, this PR adds the following to all components that currently have an $uuid proerty:

```php
public function __construct(
    public ?string $id = null, // Added this line if missing.
    ...
) {
    $this->uuid = "mary" . md5(serialize($this)) . $id; // Appended $id here if missing.
    ...
}
```

Closes #834.

### DIFF
--- a/src/View/Components/Accordion.php
+++ b/src/View/Components/Accordion.php
@@ -11,9 +11,10 @@ class Accordion extends Component
     public string $uuid;
 
     public function __construct(
+        public ?string $id = null,
         public ?bool $noJoin = false,
     ) {
-        $this->uuid = "mary" . md5(serialize($this));
+        $this->uuid = "mary" . md5(serialize($this)) . $id;
     }
 
     public function render(): View|Closure|string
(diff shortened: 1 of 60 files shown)


robsontenorio:
@anfeichtinger Phew... we have a lot of components!

Thank you!

4khobta:
Deprecation in many components
`PHP Deprecated:  Mary\View\Components\Icon::__construct(): Optional parameter $id declared before required parameter $name is implicitly treated as a required parameter in /home/www/xcrm-laravel/vendor/robsontenorio/mary/src/View/Components/Icon.php on line 15`

robsontenorio:
Cc @anfeichtinger

anfeichtinger:
#868

PR #843, opened 2025-04-02
Breadcrumbs: introduce as new component

### LINKED ISSUE #798: breadcrumbs
Will be great if also breadcrumbs will be added

https://daisyui.com/components/breadcrumbs/

anfeichtinger:
This PR introduces a new component for breadcrumbs 🚀

While the component follows the general HTML structure of daisy-ui, it offers a lot more flexibility and customization. In order to allow that the [breadcrumb classes](https://daisyui.com/components/breadcrumbs/) are not actually used.

A PR for the mary-ui.com page will follow shorty. Here are some examples:

image

image

image

Closes #798

### DIFF
--- a/src/MaryServiceProvider.php
+++ b/src/MaryServiceProvider.php
@@ -11,6 +11,7 @@ use Mary\View\Components\Accordion;
 use Mary\View\Components\Alert;
 use Mary\View\Components\Avatar;
 use Mary\View\Components\Badge;
+use Mary\View\Components\Breadcrumbs;
 use Mary\View\Components\Button;
 use Mary\View\Components\Calendar;
 use Mary\View\Components\Card;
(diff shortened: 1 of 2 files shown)


anfeichtinger:
@robsontenorio I just noticed that I forgot to use `<x-mary-icon` instead of `<x-icon`. This could lead to issues with custom prefixes so if you want to merge this today you'll have to change it yourself. Otherwise I will update the PR tomorrow.

robsontenorio:
@anfeichtinger Thanks for this! No problem, we can wait for tomorrow.

robsontenorio:
@anfeichtinger is it good to review?

anfeichtinger:
@robsontenorio Yes, you can review it now.

robsontenorio:
@anfeichtinger 

Could please check if it's fine on your side?

- Added support for small screens
- Made default style more compact to match the overall size of other components.
- Extracted tooltip logic to methods


image


image

anfeichtinger:
Two small problems (in the docs, not with the component):

- There is a typo 'seperator' in the example, leading it to not use the custom icon. And when using the m-minus custom icon it would be good to rotate it 90 degrees.
- The underline class does not work as intended inside of the mockup class. We should probably use a different example.

image

Other than that, very nice work! Thank you.

robsontenorio:
@anfeichtinger i will use other example instead of "underline", because a have some CSS on maryUI docs that prevent it shows on example. And fix `seperator`. All good, thanks again !

PR #854, opened 2025-04-08
All components: use daisyUI CSS variable for borders.

### LINKED ISSUE #851: Borders should follow daisy-ui border width
All borders in the components should follow the border size and color from the current daisy-ui theme. One instance in particular where I found a missmatch is in the Pagination component. Here is an example with the border set to 4px.

Image

This can be achieved by changing the `border` (or `border-t` and similar) to something like `border-[length:var(--border)]`. 
The colors should be consistent, something like `border-base-content/10` but I think that is the case.

I'm going to collect a few instances where we should consider updating the border size (Please note that some components have more than one occurance):
 
- [Collapse](https://github.com/robsontenorio/mary/blob/main/src/View/Components/Collapse.php#L35)
- [Colorpicker](https://github.com/robsontenorio/mary/blob/main/src/View/Components/Colorpicker.php#L88)
- [ImageLibrary](https://github.com/robsontenorio/mary/blob/main/src/View/Components/ImageLibrary.php#L156)
- [Popover](https://github.com/robsontenorio/mary/blob/main/src/View/Components/Popover.php#L59)
- [Signature](https://github.com/robsontenorio/mary/blob/main/src/View/Components/Signature.php#L87)
- [Hr](https://github.com/robsontenorio/mary/blob/main/src/View/Components/Hr.php#L32)
- [Pagination](https://github.com/robsontenorio/mary/blob/main/src/View/Components/Pagination.php#L37)
- [Spotlight](https://github.com/robsontenorio/mary/blob/main/src/View/Components/Spotlight.php#L166)
- [TimelineItem](https://github.com/robsontenorio/mary/blob/main/src/View/Components/TimelineItem.php#L32)

anfeichtinger:
This is an example of how to adhere border sizes to daisy-ui themes. Here is a quick comparison:

## Current (old) behaviour:
Default theme 1px boder:
image

Custom theme 5px border:
image

## New behaviour:
Default theme 1px boder:
image

Custom theme 5px border:
image

## Additional notes
The theme was changed by adding the following code to app.css
```css
@plugin "daisyui/theme" {
    name: light;
    --border: 5px;
}
```

I also noticed a slight difference in color in this example. The border uses `border-t-base-content/10` but the table rows use `border-t-base-content/5`, not sure if that is intentional.

Closes #851

### DIFF
--- a/src/View/Components/Card.php
+++ b/src/View/Components/Card.php
@@ -73,7 +73,7 @@ class Card extends Component
                             </div>
 
                             @if($separator)
-                                <hr class="mt-3 border-base-content/10" />
+                                <hr class="mt-3 border-t-[length:var(--border)] border-base-content/10" />
 
                                 @if($progressIndicator)
                                     <div class="h-0.5 -mt-4 mb-4">
@@ -96,7 +96,7 @@ class Card extends Component
 
                     @if($actions)
                         @if($separator)
-                            <hr class="mt-5 border-base-content/10" />
+                            <hr class="mt-5 border-t-[length:var(--border)] border-base-content/10" />
                         @else
                             <div></div>
                         @endif
--- a/src/View/Components/Collapse.php
+++ b/src/View/Components/Collapse.php
@@ -32,7 +32,7 @@ class Collapse extends Component
                 <div
                     {{
                         $attributes->class([
-                            'collapse border border-base-content/10',
+                            'collapse border-[length:var(--border)] border-base-content/10',
                             'join-item' => !$noJoin,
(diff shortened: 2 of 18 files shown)


robsontenorio:
@anfeichtinger 

This looks good! Are you able to change some more components in this same PR for now? If not, we can merge this.

anfeichtinger:
@robsontenorio I changed it everywhere where it is easy to do, please review carefully as there are a lot of affected components.

For Timeline it cannot be easily done as it relies heavily on fixed values.

robsontenorio:
@anfeichtinger Perfect! 10/10

It was fun to see the entire docs within `border: 5px` 😄

PR #856, opened 2025-04-08
Password: fix button spacing and hover effect

### LINKED ISSUE #849: Password: Inconsistent icon size and strange hover effect
### maryUI version

2.0.1

### daisyUI version

5.0.9

### Livewire version

3.6.2

Chrome

### What happened?

When having a normal input and a password field side by side (e. g. at a login page), I noticed that the leading icon does not take up the same amount of space.

Code from my project:
```blade
{{-- Email --}}
<x-input type="email" icon="o-envelope" :label="__('labels.email')" class="w-full" wire:model='email'
    omit-error />

{{-- Password --}}
<x-password :label="__('labels.password')" wire:model='password' omit-error />
```

Screenshot (the blueish background comes from autofill and makes the spacing more noticable.
Image

Also the icon button is getting a hover effect whereever the input is hovered. So even if I hover the input on the far right, the icon gets the effect.

anfeichtinger:
This will fix both:

- the inconsistent spacing of the visibility toggle button by adding a ``-m-1` class.
- activating the hover effect on the toggle button when hovering the input somewhere else.
    -  This is done by changing the `label` tags to `div`. Note: I am not sure why we are using two nested `label` tags anyways. Is there an explanation for this? @robsontenorio 

Closes #849

### DIFF
--- a/src/View/Components/Password.php
+++ b/src/View/Components/Password.php
@@ -93,7 +93,7 @@ class Password extends Component
                         </legend>
                     @endif
 
-                    <label @class(["floating-label" => $label && $inline])>
+                    <div @class(["floating-label" => $label && $inline])>
                         {{-- FLOATING LABEL--}}
                         @if ($label && $inline)
                             <span class="font-semibold">{{ $label }}</span>
@@ -106,7 +106,7 @@ class Password extends Component
                             @endif
 
                             {{-- THE LABEL THAT HOLDS THE INPUT --}}
-                            <label
+                            <div
                                 x-data="{ hidden: true }"
 
                                 {{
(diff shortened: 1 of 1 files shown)


robsontenorio:
The nested labels is for default vs inline style.

robsontenorio:
@anfeichtinger All inputs field follows this "label" pattern, like daisyUI. But, I will leave as is because of this special case of clickable button.

You will see this same "issue" on input and select slot examples on docs. For these we will need find another solution.

anfeichtinger:
> All inputs field follows this "label" pattern, like daisyUI.

But does the label actually have a use in the structure? Would the icon not be clickable otherwise or are there any visual or accessibility issues? I would expect something like [this](https://daisyui.com/components/input/#with-fieldset-and-fieldset-legend) but I guess the components here are more complex than that.

I only found the label pattern in [examples with validation](https://daisyui.com/components/input/#email-input-with-icon-validator-button-join) but we use something different for validation errors anyways.

I couldn't find another solution with css for now, it might only be "fixable" by changing to a different tag because that's just how labels work. 😅

PR #862, opened 2025-04-13
Rich Text Editor: automatic theme switching (dark/light)

robsontenorio:
Add the ability to automatically switch the theme (dark/light).

It is important to know it toggles between the own internal TinyMCE  themes ('oxide' and 'oxide-dark') based on dark variant class. TinyMCE **will not** follow custom themes, but will use its own variant.

Hey guys, could please check if everything is fine on your side with this branch?
Mostly, check if some content or state will be lost when toggling theme.

 @amnesca @sergeynilov @ekrist1 @alexnogueirasilva

```
composer require robsontenorio/mary:dev-editor-theme
php artisan view:clear
```

### DIFF
--- a/src/View/Components/Editor.php
+++ b/src/View/Components/Editor.php
@@ -90,7 +90,7 @@ class Editor extends Component
                                     uploadUrl: '{{ $uploadUrl }}?disk={{ $disk }}&folder={{ $folder }}&_token={{ csrf_token() }}'
                                 }"
                             x-init="
-                                tinymce.init({
+                                let settings = {
                                     {{ $setup() }},
 
                                     @if($gplLicense)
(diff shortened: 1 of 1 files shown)


amnesca:
Hi Robson,
The first issue is that the following event listener should be 'theme-changed' not 'theme-changed-class':

Editor.php
```js
window.addEventListener('theme-changed', async (e) => {
    settings.skin = e.detail == 'dark' ? 'oxide-dark' : 'oxide'
    settings.content_css = e.detail == 'dark' ? 'dark' : 'default'
    tinymce.get('{{ $id ?? $uuid }}').destroy()
    tinymce.init(settings)
})
```

And, as previously discussed, to enable the use of dark themes other that 'dark', the variant should be determined using the class instead of the theme attribute. So the $dispatch in the 'ThemeToggle' should send the class:

ThemeToggle.php
```js
setToggle() {
    document.documentElement.setAttribute('data-theme', this.theme)
    document.documentElement.setAttribute('class', this.class)
    this.$dispatch('theme-changed', this.class)
},
```

Also, I'm still doing some other testing, but it seems like destroying and reinitialising the TinyMCE editor completely wipes the undo manager, so perhaps this should be an optional feature, or at least explained in the docs to mitigate any confusion?

amnesca:
Also, changing to pages that don't have an editor produces the following console error:

Screenshot from 2025-04-14 17-59-14

amnesca:
I agree, this is one of the reasons we're currently looking into the [Tiptap editor](https://github.com/ueberdosis/tiptap) instead. It also has some really powerful customisations through extensions too.

PR #863, opened 2025-04-15
Editor: fix TinyMCE undo/redo actions and dark mode setting

amnesca:
1. Undo/redo actions are not committing to the editor content and require separate event listeners in the editor setup to recognise these actions.
2. The editor receives a dark variant from the 'data-theme' attribute. This works fine if you only use the 'dark' theme name to toggle dark mode, however, it won't work if you use another dark theme name. The solution is to change TinyMCE's 'skin' and 'content_css' settings to accept the 'dark' variant from the 'class' instead of the 'data-theme' attribute.


robsontenorio:
@amnesca  Thanks!

PR #867, opened 2025-04-16
Add icon and icon classes support for Header

lchevalot:
Added the `icon` property to let the user pick an icon for the header, along with an `icon-classes` property to let the user have control of the size, margin, colors, ... of the icon.

Added some flex classes to the title parent `<div>` to align the title and the icon automatically, and wrapped the title content in a `<span>` to add a small margin when there's an icon.

Showcase of the result :

image

image

```html
<x-header icon="o-gif" icon-classes="w-8 h-8 text-red-500" title="Hello" subtitle="messing around with the icon" separator progress-indicator>
        <x-slot:middle class="!justify-end">
            <x-input placeholder="Search..." wire:model.live.debounce="search" clearable icon="o-magnifying-glass" />
        </x-slot:middle>
        <x-slot:actions>
            <x-button label="Filters" @click="$wire.drawer = true" responsive icon="o-funnel" class="btn-primary" />
        </x-slot:actions>
    </x-header>
```

### DIFF
--- a/src/View/Components/Header.php
+++ b/src/View/Components/Header.php
@@ -19,6 +19,10 @@ class Header extends Component
         public ?bool $withAnchor = false,
         public ?string $size = 'text-2xl',
 
+        // Icon
+        public ?string $icon = null,
+        public ?string $iconClasses = null,
+
         // Slots
         public mixed $middle = null,
         public mixed $actions = null,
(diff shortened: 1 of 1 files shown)


robsontenorio:
@lchevalot  Perfect! Thanks!

Issue #870, opened 2025-04-17
DispatchChangeEvent doesn't call in selectAll method of Choices component


abduraim:
### maryUI version

2.2.2

### daisyUI version

5.0.18

### Livewire version

3.6.2

### What browsers are affected?

Chrome

### What happened?

Good day. I encountered the following problem: when I click on the item "select all", the event does not trigger.

Need add call to selectAll() method of Choices component.

**_Choices.php (148 line)_**

```php
                            selectAll() {
                                this.selection = this.options.map(i => i.{{ $optionValue }})
                            },
```

```php
                            selectAll() {
                                this.selection = this.options.map(i => i.{{ $optionValue }})
                                +++ this.dispatchChangeEvent({ value: this.selection})
                            },
```

I wrote PR - https://github.com/robsontenorio/mary/pull/871

Thanks

PR #871, opened 2025-04-17
fix: add dispatchChangeEvent call to selectAll method of Choices comp…

abduraim:
Resolve problem of [this issue](https://github.com/robsontenorio/mary/issues/870)


robsontenorio:
Thanks @abduraim !

Issue #853, opened 2025-04-08
Cannot set choices component height with daisyUI select classes


lchevalot:
### maryUI version

2.0.1

### daisyUI version

5.0.16

### Livewire version

3.6.2

### What browsers are affected?

Chrome

### What happened?

Before the MaryUI 2 update, it was possible to change the `<x-choices />` component height with DaisyUI select classes (**select-sm**, **select-md**, **select-lg**, ...). Since the update, it is no longer possible (while using **select-primary** or other color changes still work on the component).

I've tried to mess around in the DOM and it seems to be related to the **h-fit** class in the label of the component (_cf. screenshot_). When removed the **select-\*size\*** classes work normally.

![Image](https://github.com/user-attachments/assets/416600b3-9c84-4f5a-9f74-495498a614ac)

I don't know if that is intended. I can submit a PR where i remove the **h-fit** or replace it with **h-max** if needed.

lchevalot:
Furthermore, the **h-fit** also messes up the alignement between the input and some joined items on specific themes (pastel, caramellatte, ... Ones with thick borders on inputs in general) (_cf. screenshot_)

![Image](https://github.com/user-attachments/assets/41b5d77f-27e4-4750-9c14-bf33789421d7)

Alignement goes back to normal when **h-fit** is removed.

![Image](https://github.com/user-attachments/assets/190974c2-75e7-4cac-b036-1218d5f3c0c0)

lchevalot:
@robsontenorio I've tried to play with the DOM, this is a pain in the ass for sure lol.

I've tried to set the parent div of the choice component as **h-full**, aswell with the choice component in **h-full** itself instead of **h-fit**. This works as you can now use **select-*size*** classes and the component still grows up vertically if needed. In both cases, the mobile version is a bit broken because the crosses in the badges inside the select are out of box.

https://www.loom.com/share/8ee4ca4657b445e58ab318dc9957dbe1?sid=197bd5bb-72c4-4e89-9dd1-59128eda7c93

We could also juste remove the height classes and make the overflow content of the choice scrollable. I don't see any other solution yet.

robsontenorio:
@lchevalot could  you please send a Draft PR with this change ? So , I can play around with it.

PR #873, opened 2025-04-17
Choices: fix select height (#853)

lchevalot:
@robsontenorio Turns out it was way more simpler than we thought (or at least i hope so).

I hope this suits you.


robsontenorio:
@lchevalot this solve/ the issue for “select-<size>” and also for small screen auto grown height ?

robsontenorio:
Thanks @lchevalot !

The Choices component is the most complex ever.

robsontenorio:
Thanks @lchevalot !

PR #880, opened 2025-04-23
MenuItem: Add disabled property and rename enabled as hidden

lchevalot:
This PR renames the `$enabled` property as `$hidden` for better readability, and adds the `$disabled` property to allow a menu item to be disabled based on the [daisyUI menu disabled class](https://daisyui.com/components/menu/#menu-with-disabled-items).

Showcase of the result :

image

image

Code sample :

```blade
<x-menu activate-by-route>
    <x-menu-item title="Hello" icon="o-sparkles" link="/" />

    <x-menu-sub title="Settings" icon="o-cog-6-tooth">
        <x-menu-item :disabled=true title="Wifi" icon="o-wifi" link="####" />
        <x-menu-item :hidden=true title="Archives" icon="o-archive-box" link="####" />
    </x-menu-sub>
</x-menu>
```

### DIFF
--- a/src/View/Components/MenuItem.php
+++ b/src/View/Components/MenuItem.php
@@ -25,7 +25,8 @@ class MenuItem extends Component
         public ?string $badgeClasses = null,
         public ?bool $active = false,
         public ?bool $separator = false,
-        public ?bool $enabled = true,
+        public ?bool $hidden = false,
+        public ?bool $disabled = false,
         public ?bool $exact = false
     ) {
         $this->uuid = "mary" . md5(serialize($this)) . $id;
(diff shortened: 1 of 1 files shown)


anfeichtinger:
While I agree that renaming the `$enabled` property makes sense and is easier to understand, this would be considered a breaking change as everyone using it needs to change their code.

lchevalot:
@anfeichtinger my fault, i didn't understand that breaking change meant that i have to adapt to people already using it.

```php
if ($this->hidden === true || $this->enabled === false)
``` 

I suppose i should go with that ?

robsontenorio:
This is a simple one, in favor of a better consistency in naming. I think it worth it.

Thanks @lchevalot

PR #881, opened 2025-04-24
Carousel: Added optional autoplay setting to carousels (#878)

### LINKED ISSUE #878: Carousel autoplay
A variable to get a carousel to autoplay would be a nice feature.
        
It probably does not need much more than this addition to the `x-data` of the carousel's `div`:
        
```blade
init() {
    setInterval(() => { this.next(); }, 2000 /* Make this a parameter of the timer? */);
}
```

 - It would probably be a good idea to enforce some minimum value for the timer.
 - By default or if the timer is set to a `<= 0` value, the carousel does not autoplay.

AtmoFX:
Added optional variables `autoplay` and `interval` to control the carousel behavior, respectively `false` and `2000`ms by default.

Note: an `autoplay` carousel with too short and `interval` (typically under 100ms but that may depend on the device / browser) may look broken.

Close #878

### DIFF
--- a/src/View/Components/Carousel.php
+++ b/src/View/Components/Carousel.php
@@ -19,6 +19,8 @@ class Carousel extends Component
         public ?string $id = null,
         public ?bool $withoutIndicators = false,
         public ?bool $withoutArrows = false,
+        public ?bool $autoplay = false,
+        public ?int $interval = 2000,
 
         // Slots
         public mixed $content = null,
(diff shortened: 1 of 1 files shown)


src/View/Components/Carousel.php
```diff
@@ -19,6 +19,8 @@ public function __construct(
         public ?string $id = null,
         public ?bool $withoutIndicators = false,
         public ?bool $withoutArrows = false,
+        public ?bool $autoplay = false,
```
themegazord, line 22: I guess it's not necessary declare this variable with null, because receive ever only 2 value being a boolean variable.

robsontenorio:
@AtmoFX Perfect! Thanks!

PR #886, opened 2025-04-29
MenuSub: Add disabled property and rename enabled as hidden

lchevalot:
In addition of #880, it works the same way.


robsontenorio:
Thanks @lchevalot !

PR #888, opened 2025-04-29
Choices/ChoicesOffline: fix badge height #879

### LINKED ISSUE #879: Choice/Choice-Offline : badge on label does not fill the area
### maryUI version

2.2.3

### daisyUI version

5.0.9

### Livewire version

3.5.20

Microsoft Edge

### What happened?

Hi,

Choice/Choice-Offline : badge on label does not fill the area if the label is longer than the choice area, as in the picture below :

Image

I think this is because the badge component has a fixed height attribute

Image

Can I add `!h-auto` to the badge component to fix this?

Or is there a better way?

If I may I will make a PR to add `!h-auto` to the following code

```
<span class="mary-choices-element cursor-pointer badge badge-soft m-0.5 !inline-block !h-auto">
```

Thanks

akhmads:
fix #879 

Choice and choice offline badge height 

Preview (Before) :

image

Preview (After) :

image

Thank You

### DIFF
--- a/src/View/Components/Choices.php
+++ b/src/View/Components/Choices.php
@@ -287,7 +287,7 @@ class Choices extends Component
                                                     </div>
                                                 @else
                                                     <template x-for="(option, index) in selectedOptions" :key="index">
-                                                        <span class="mary-choices-element cursor-pointer badge badge-soft m-0.5 !inline-block">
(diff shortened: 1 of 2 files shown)


robsontenorio:
Thanks @akhmads

PR #889, opened 2025-04-30
Table: add `footer` slot

akhmads:
I want to add a slot for table footer with code example like below, is it possible or any better idea?

```
<x-table>
    <x-slot:footer class="bg-gray-50 dark:bg-gray-800">
        <tr>
            <td class="text-right font-bold">Total</td>
            <td class="text-right font-bold">1,200,00</td>
            <td></td>
            <td></td>
            <td></td>
        </tr>
    </x-slot:footer>
</x-table>
```

image

Thank you


robsontenorio:
Thanks @akhmads !

Issue #848, opened 2025-04-07
Extra `<div>` in `card` title


AtmoFX:
I am trying to put 2 `div`s side by side in a card title and see some extra `<div>` are inserted for the title tag.
That behavior is different when compared to the figure slot and I could not figure out a benefit for this difference.

Here is an example to compare the figure and the title slots:

```
<x-card shadow class="border border-gray-400/20">
    <x-slot:figure class="m-0 flex !justify-between w-full mark-the-figure">
        <div>ABC</div>
        <div>DEF</div>
    </x-slot:figure>
    <x-slot:title class="flex !justify-between w-full mark-the-title">
        <div>GHI</div>
        <div>JKL</div>
    </x-slot:title>
    Card text
</x-card>
```
The `mark-the-figure` and `mark-the-title` classes do nothing other than help track where each item is rendered in the DOM.

This results in only the figure between correctly applied the `!justify-between` class (or to be more precise, this is what it looks like; the actual explanation is written below):

![Image](https://github.com/user-attachments/assets/0cf2ef55-ce24-48c2-baa5-7e7e79a1d23f)

The reason can be found in the generated DOM:
```
<div class="card bg-base-100 rounded-lg p-5 shadow-xs border border-gray-400/20" wire:key="mary5dc6f36fcd268db20f5fce81cad760c2">
    <figure class="mb-5 -m-5 m-0 flex !justify-between w-full mark-the-figure">
        <div>ABC</div>
        <div>DEF</div>
    </figure>
    <div class="pb-5">
        <div class="flex justify-between items-center"> <!-- This div ... -->
            <div> <!-- ... and this one -->
                <div class="text-xl font-bold flex !justify-between w-full mark-the-title">
                    <div>GHI</div>
                    <div>JKL</div>
                </div>
            </div>
        </div>
    </div>
    <div>
        Card text
    </div>
</div>
```
As a result, there is no way to make the title take all the available width (hence the reason why it looks like `!justify-between` has no effect).

Had the generated HTML been like below (without the 2 commented `<div>`), the title would have spanned across the whole card width:

```
<div class="card bg-base-100 rounded-lg p-5 shadow-xs border border-gray-400/20" wire:key="mary5dc6f36fcd268db20f5fce81cad760c2">
    <figure class="mb-5 -m-5 m-0 flex !justify-between w-full">
        <div>ABC</div>
        <div>DEF</div>
    </figure>
    <div class="pb-5">
[...]

PR #892, opened 2025-05-02
Card: Layout improvements

AtmoFX:
Improvements in this commit are:
 1. From the original question in issue #848: The title and subtitle will try to fill the full width of the card.<br/>This allows the title and subtitle to be divided into 2 parts (e.g. the title text and an icon), clearly separted from one another by e.g. applying `justify-between` to it.<br/>FYI: I have introduced a `gap-3` (same as the gap applied to the actions slot) between the title/subtitle and the menu. The card looks off without it.
 2. From my additional question in issue #848: The actions will also fill the full width of the card. Moreover, the actions slot can now be specified CS `class`es.<br/>This allows to lay them out in distrinct groups too.<br/>Note that unlike the title/subtitle, actions are applied `justify-end`. As a result, `justify-between` alone will not work; `!justify-between` must be used instead to override it. You should probably include that in your documentation. Alternatively, you could remove the default `justify-end` so that control of the button placement is easier to users, but I am not going to make that call myself. 
 3. The default slot is now applied `grow-1`.<br/>This pushes the actions and the separator between them as far down as possible, when, like in the below screenshot, cards are layed out in a CSS grid and do not have the same intrinsic height.<br/>On that topic, I added `items-end` to the actions slot, so that even if it is made of several `div`s, each of a different height, buttons will preferentially be aligned to the bottom of the padding area.<br/>Note that the  buttons in the 2 _Your stats_  cards are in the default slot, **not** in the actions slot, explaining why the `Save` buttons stay up there. 

Below is a of the improvements.
 - First are the 4 cards showcased on https://mary-ui.com/docs/components/card, hopefully showing this does not break anything down.
 - The last 2 cards in the grid show the 3 above improvements.

```
<x-card separator>
    Full width title + full width actions (using spacer div)
    <x-slot:figure>
        <img src="https://picsum.photos/500/200" />
    </x-slot:figure>
    <x-slot:title class="flex flex-row justify-between w-full">
        <div>ABC</div>
        <div>DEF</div>
    </x-slot:title>
    <x-slot:subtitle class="flex flex-row justify-between w-full">
        <div>GHI</div>
        <div>JKL</div>
    </x-slot:subtitle>
    <x-slot:actions separator class="flex !justify-between">
        <x-button label="Action 1" class="btn-secondary" />
        <x-button label="Action 2" class="btn-secondary" />
        <div class="grow-1"></div>
        <x-button label="Ok" class="btn-primary" />
    </x-slot:actions>
</x-card>

<x-card separator>
    Full width title (minus menu) + full width actions (using child div) 
    <x-slot:figure>
        <img src="https://picsum.photos/500/200" />
    </x-slot:figure>
    <x-slot:title class="flex flex-row !justify-between w-full">
        <div>ABC</div>
        <div>DEF</div>
    </x-slot:title>
    <x-slot:subtitle class="flex flex-row !justify-between w-full">
        <div>GHI</div>
        <div>JKL</div>
    </x-slot:subtitle>
    <x-slot:menu>
        <x-button icon="o-share" class="btn-circle btn-sm" />
        <x-icon name="o-heart" class="cursor-pointer" />
    </x-slot:menu>
    <x-slot:actions separator class="flex !justify-between">
        <div class="flex gap-3">
        <x-button label="Action 1" class="btn-secondary" />
        <x-button label="Action 2" class="btn-secondary" />
        </div>
        <x-button label="Ok" class="btn-primary" />
    </x-slot:actions>
</x-card>
```

image

### DIFF
--- a/src/View/Components/Card.php
+++ b/src/View/Components/Card.php
@@ -53,7 +53,7 @@ class Card extends Component
 
                     @if($title || $subtitle)
                         <div class="pb-5">
-                            <div class="flex justify-between items-center w-full">
+                            <div class="flex gap-3 justify-between items-center w-full">
                                 <div class="grow-1">
                                     @if($title)
                                         <div @class(["text-xl font-bold", is_string($title) ? '' : $title?->attributes->get('class') ]) >
--- a/src/View/Components/Card.php
+++ b/src/View/Components/Card.php
@@ -53,8 +53,8 @@ class Card extends Component
 
                     @if($title || $subtitle)
                         <div class="pb-5">
-                            <div class="flex justify-between items-center">
-                                <div>
+                            <div class="flex justify-between items-center w-full">
+                                <div class="grow-1">
                                     @if($title)
                                         <div @class(["text-xl font-bold", is_string($title) ? '' : $title?->attributes->get('class') ]) >
                                             {{ $title }}
@@ -90,7 +90,7 @@ class Card extends Component
                         </div>
                     @endif
 
-                    <div>
+                    <div class="grow-1">
                         {{ $slot }}
                     </div>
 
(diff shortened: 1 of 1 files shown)


robsontenorio:
@AtmoFX Great job! Thanks!

PR #893, opened 2025-05-02
Fix header loading progress

akhmads:
Hello,

I found a difference in loading in the Header compared to Card and HR.

Before PR (Header) :

Header Progress Before

After PR (Header) :

Header Progress After

I just changed the height of the progress bar. Hopefully this can be accepted because the progress display is not the same as the others (Card, HR).

Thank You

### DIFF
--- a/src/View/Components/Card.php
+++ b/src/View/Components/Card.php
@@ -53,8 +53,8 @@ class Card extends Component
 
                     @if($title || $subtitle)
                         <div class="pb-5">
-                            <div class="flex justify-between items-center">
-                                <div>
+                            <div class="flex gap-3 justify-between items-center w-full">
+                                <div class="grow-1">
                                     @if($title)
                                         <div @class(["text-xl font-bold", is_string($title) ? '' : $title?->attributes->get('class') ]) >
                                             {{ $title }}
@@ -90,7 +90,7 @@ class Card extends Component
                         </div>
                     @endif
 
-                    <div>
+                    <div class="grow-1">
                         {{ $slot }}
                     </div>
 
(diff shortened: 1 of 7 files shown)


reafeichtinger:
I don't think this is the right solution. The Progress bar uses the height of the daisy-ui theme, you might want to adjust that instead. If there is an inconsistency then it should be fixed by updating the `hr` to match. If it is necessary to still customize the height maybe we should add a property for that instead. What do you think @robsontenorio?

akhmads:
@reafeichtinger 

I think so too, for Hr it would be better if these components could be equated from one component.

I just only see height difference between Card.php, Hr.php and Header.php

Card.php

```
<progress
	class="progress progress-primary w-full h-0.5"
	wire:loading

	@if($progressTarget())
		wire:target="{{ $progressTarget() }}"
	@endif></progress>
```

Hr.php

```
<progress
	class="progress progress-primary hidden h-[1px]"
	wire:loading.class="!h-[length:var(--border)] !block"

	@if($progressTarget())
		wire:target="{{ $progressTarget() }}"
	@endif></progress>
```

Header.php

```
<progress
	class="progress progress-primary w-full h-[var(--border)]"
	wire:loading

	@if($progressTarget())
		wire:target="{{ $progressTarget() }}"
	@endif></progress>
```

So I just made the height the same between the three components.

If there is a better solution, I will be happy

reafeichtinger:
If anything, all components then should use `h-[var(--border)`. It needs to be consistent and fit together well.

robsontenorio:
@akhmads ,

Previously we had a hardcoded height. But , @reafeichtinger proposed to use var instead, so we can use daisyUI themes properly.

akhmads:
Noted sir

robsontenorio:
@akhmads  To be honest, I wouldn't be bothered by this minor issue. Most users probably won't even notice it.

Issue #896, opened 2025-05-05
Issue with prefixed mary ui when using header


Scraayp:
### maryUI version

2.3

### daisyUI version

5.0.17

### Livewire version

3.0

### What browsers are affected?

Chrome

## HTML/Blade
```blade
<x-mary-header title="Gearchiveerde projecten" icon="o-archive-box" subtitle="Alle gearchiveerde projecten met een mogelijkheid om ze terugtestellen." separator
		class="mx-auto w-[80%]" />
```
## Related Code:
https://github.com/robsontenorio/mary/blob/263e483874a1c303173de51f9c288624e031e68b/src/View/Components/Header.php#L54

## Error:
Svg by name "o-archive-box" from set "default" not found.

I have my mary-ui prefixed with mary so that all the components work with x-mary-*. This has been working on every component. Just when header with icon, I get the problem that it's trying to find a icon with x-icon while it should be x-mary-icon. 

## Best Solution:
Get the icon internally and not call a component. This would fix the issue with the prefix.

_🔥Point of notice: I really appreciate all the effort you’ve poured into the package—it looks fantastic, and I’ve already integrated it into several of my own applications._

reafeichtinger:
This can be easily fixed by changing the `<x-icon` to `<x-mary-icon` in the Header component. A PR is welcome.

Scraayp:
Done!

robsontenorio:
Closed by #898

PR #898, opened 2025-05-07
Header: use the internal "x-mary-icon" component

Scraayp:
#### 🐛  The problem  
When a project changes Mary’s default component prefix (e.g. from `<x-*>` to `<x-mary-*>`), the `<x-mary-header>` component still tried to render its leading icon with:

```blade
<x-icon name="{{ $icon }}" … />
```
Because the tag itself wasn’t prefixed, Livewire/Laravel couldn’t resolve the component class and the icon simply disappeared (see issue #896).

#### ✅  What this PR does
File	Change
```
src/View/Components/Header.php	Replace <x-icon …> with <x-mary-icon …>
```
Only a single line is touched; behaviour is otherwise unchanged.

### DIFF
--- a/src/View/Components/Header.php
+++ b/src/View/Components/Header.php
@@ -51,7 +51,7 @@ class Header extends Component
                                 @endif
                                 
                                 @if($icon)
-                                    <x-icon name="{{ $icon }}" class="{{ $iconClasses }}" />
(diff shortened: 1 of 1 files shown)


robsontenorio:
Thanks!

PR #907, opened 2025-05-20
Table: change header sort icons and add neutral sort state

mani-damirchilou:
1. Replace existing `o-arrow-small-down` and `o-arrow-small-up` table sort icons with `o-chevron-down` and `o-chevron-up`.
2. Add neutral state for distinguishing which headers are sortable and which are not.


robsontenorio:
Thanks @mani-damirchilou !

PR #908, opened 2025-05-20
Choices/ChoicesOffline: handle dynamic `disabled` and `readonly` states

### LINKED ISSUE #899: Disabled Choiches are still usable
### maryUI version

2.3.0

### daisyUI version

5.0.35

### Livewire version

v3.6.3

Chrome, Firefox

### What happened?

I saw there was a bug reported one year ago so I think the issue came back after the upgrade to mary v2.

As you can see even after disabling the choice it can still be clicked.

https://github.com/user-attachments/assets/3ea4b2c6-972c-4bbd-9d38-601a7d91ad4d

HTML/Blade

```                <x-choices
                    inline
                    label="Organization"
                    wire:model="organization_id"
                    search-function="organizations"
                    :options="$organizationsSearch"
                    searchable
                    single
                    clearable
                    :disabled="!$categories[$category_id]['organizations']"
                ></x-choices>
                <x-choices
                    class="join-item"
                    inline
                    label="User"
                    wire:model="user_id"
                    search-function="users"
                    :options="$userSearch"
                    searchable
                    single
                    clearable
                    :disabled="!$categories[$category_id]['users']"
                >
                    @scope("item", $user)
                        <x-list-item :item="$user" sub-value="email" class=" {{$user->role == App\RoleType::ADMIN->value ? 'bg-error/10': ''}}"/>
                    @endscope

                    {{-- Selection slot --}}
                    @scope("selection", $user)
                        {{ $user->name }}
                    @endscope
                </x-choices>
```

This is the code for the two choices, I don't think there is anything wrong unless I missed something.

mani-damirchilou:
Fix #899 

Disabled `<x-choices>` component is no longer clickable & focusable and it can be reactive based on the value provided by the parent livewire component.

### DIFF
--- a/src/View/Components/Choices.php
+++ b/src/View/Components/Choices.php
@@ -55,7 +55,7 @@ class Choices extends Component
         public mixed $prepend = null,
         public mixed $append = null
     ) {
-        $this->uuid = "mary" . md5(serialize($this)) . now()->timestamp;
+        $this->uuid = "mary" . md5(serialize($this)) . $id;
 
         if (($this->allowAll || $this->compact) && ($this->single || $this->searchable)) {
             throw new Exception("`allow-all` and `compact` does not work combined with `single` or `searchable`.");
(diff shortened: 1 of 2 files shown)


robsontenorio:
Thanks @mani-damirchilou,

However, we cannot use "wire:key" on alpineJS divs, as the internal state will change after each keystroke.

I have updated the implementation to utilize `$isDisabled()` to ensure the component is properly disabled.

PR #910, opened 2025-05-24
Header: add `progress-indicator-class`

### LINKED ISSUE #906: <x-header> progress bar color.
Hello there, 

I really love mary-ui and I love to integrate this in my projects. 💖

I just have a question about the `<x-header>` progress bar. Is there a way to change the progress bar color without customizing the theme? Or if ever, if there's a way to use a slot for the progress bar to use the `<x-loading />` there?

Thank you.

themegazord:
Fix #906

### DIFF
--- a/src/View/Components/Header.php
+++ b/src/View/Components/Header.php
@@ -16,6 +16,7 @@ class Header extends Component
         public ?string $subtitle = null,
         public ?bool $separator = false,
         public ?string $progressIndicator = null,
+        public ?string $progressIndicatorColor = "progress-primary",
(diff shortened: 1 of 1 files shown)


themegazord:
Ok, in this moment I have a problem with my internet provider, when this problem finished I see this.

robsontenorio:
Let’s keep this PR only for Header. It seems good.

robsontenorio:
Thanks @themegazord 

Just changed to `progress-indicator-class` to be consistent in naming.

PR #913, opened 2025-05-28
Input: handle `.blur` for money

marcelcunha:
When using the input component with money and locale attributes, it loses the formatting if you use `wire:model.blur`.

```php
 <x-input wire:model.blur="money" money locale="pt-BR"/>
```
image

This PR fixes the problem.

### DIFF
--- a/src/View/Components/Input.php
+++ b/src/View/Components/Input.php
@@ -152,7 +152,7 @@ class Input extends Component
                                         {{
                                             $attributes
                                                 ->merge(['type' => 'text'])
-                                                ->except($money ? ['wire:model', 'wire:model.live'] : '')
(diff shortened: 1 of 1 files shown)


robsontenorio:
I don't get why you need to use `wire:model.blur="money"` on your example. 

This would be the correct usage. It keeps the input formatted all the time. Please, check the docs live examples.

```blade
 <x-input wire:model="money" money locale="pt-BR"/>
```

marcelcunha:
The previous example is just to demonstrate what happens when using the `.blur`. 

In my application, i need to enable  some fields after the monetary field has been filled in.

robsontenorio:
Thanks @marcelcunha !

Issue #915, opened 2025-05-29
Popover: Transition glitch


amnesca:
### maryUI version

2.4.1

### daisyUI version

5.0.27

### Livewire version

3.6.3

### What browsers are affected?

Firefox, Chrome

### What happened?

Hi, I noticed that the popover component has a jumpy transition glitch when using longer content. It seems like the transition is starting before the anchor position is set properly.

![Image](https://github.com/user-attachments/assets/13693a44-d3c7-40d4-8df9-f0f73e2fd543)
 
```blade
<x-popover position="top-start" offset="10">
    <x-slot:trigger>
        <x-icon name="o-information-circle" class="w-6 h-6 hover:text-primary" />
    </x-slot:trigger>

    <x-slot:content class="prose text-xs p-6 max-w-4xl">
        {!! 
          "<p>Lorem ipsum dolor sit amet, <a href=''>consectetur</a> adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
   Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure
   dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non
   proident, sunt in culpa qui officia deserunt mollit anim id est laborum.</p>"
        !!}
    </x-slot:content>
</x-popover>
```
However, adding `x-transition` to the content div seems to remove this glitch and also provides a smoother transition compared to the current version, which can be quite abrupt when displaying larger popovers. Thanks.

**Potential Solution:**
```blade
<!-- CONTENT -->
<div
  x-show="open"
  x-anchor.{{ $position }}.offset.{{ $offset }}="$refs.myTrigger"
  x-transition
  @mouseover="show()"
  @mouseout="hide()"
  {{ $content->attributes->class(["z-[1] shadow-xl border-[length:var(--border)] border-base-content/10 w-fit p-3 rounded-md bg-base-100"]) }}
>
  {{ $content }}
</div>
```

PR #919, opened 2025-05-29
Popover: fix transition glitch

amnesca:
Added `x-transition` to popover content div to fix jumpy transition glitch when using larger popovers (refer [Issue #915](https://github.com/robsontenorio/mary/issues/915)).


robsontenorio:
Thanks @amnesca !

Issue #901, opened 2025-05-14
Font issues with title of modals


AtmoFX:
### maryUI version

2.4.0

### daisyUI version

5.0.35

### Livewire version

3.6.3

### What browsers are affected?

Firefox, Chrome, Safari, Microsoft Edge

### What happened?

DaisyUI [modals](https://daisyui.com/components/modal/) all have their titles set with `text-lg font-bold`.

MaryUI does things differently, defining the title as:

```
@if($title)
    <x-mary-header :title="$title" :subtitle="$subtitle" size="text-xl" :separator="$separator" class="!mb-5" />
@endif
```
where apparently, `x-mary-header` is used ONLY for modals (unless my VScode has somehow missed other instances of it), and forces the font with a slightly odd code:

```
<div @class(["flex", "items-center", "$size font-extrabold", is_string($title) ? '' : $title?->attributes->get('class') ]) >
```

So both the size and the weight do not match DaisyUI but on top of that, this unique header makes it difficult to tweak.

Should the header be dropped completely? Or set to accept classes passed to the modal's `title` slot (and have defaults that match DaisyUI)?

AtmoFX:
FYI, I'm still thinking about how exactly to do it.

lchevalot:
@robsontenorio I've been thinking about this, in my opinion the best way to do that would be to have a regular `title` attribute by default like this :

```blade
@if($title)
    <h3 class="font-bold text-lg mb-5">{{ $title }}</h3>
@endif
```

then if the user wants a custom title, he could uses a **header slot**, that could be implemented that way :

```php
...
// Slots
public mixed $header = null,
...
```
```blade
@if($header)
    <div>
        {{ $header }}
    </div>
@elseif($title)
    <h3 class="font-bold text-lg mb-5">{{ $title }}</h3>
@endif
```

That would also means removing the `$subtitle` attribute (unless you wanna keep it).

Then the result would be to be able to have a modal title with 2 ways :

1. The native `title` attribute :
```blade
<x-modal wire:model="myModal1" title="This is the title :)" class="backdrop-blur">
    Press `ESC`, click outside or click `CANCEL` to close.

    <x-slot:actions>
        <x-button label="Cancel" @click="$wire.myModal1 = false" />
    </x-slot:actions>
</x-modal>
```

<img width="624" height="308" alt="Image" src="https://github.com/user-attachments/assets/020e3952-2aec-4eba-a58d-d56a6cba2efe" />

2. The header slot :
```blade
[...]

PR #920, opened 2025-05-31
Fixed styling for input (regular and money)

AtmoFX:
When trying to enforce CSS classes to inputs (you will find a snippet below), intending to render `disabled` or `readonly` inputs like normal, this is what I get:

image

There are 2 issues, the first of which being about the cursor (not captured on the screenshot):
 - When disabled, the cursor is expected to be `forbidden`, as per the following 2 rules:
     ```css
     /* On the input itself */
     @layer utilities {
      .input {
        &:has(> input[disabled]) > input[disabled] {
            cursor: not-allowed;
          }
        }
      }
     /* Inherited from label */
      @layer utilities {
        .input {
          &:has(> input[disabled]), &:is(:disabled, [disabled]) {
            cursor: not-allowed;
            color: var(--color-base-content);
          }
        }
      }
     ```
     But `money` inputs have an extra `div` right above `input` in the DOM, so the cursor is only correct on the outside perimeter of what looks like the `input` (but is actually the `label`). The screenshot makes the area where the cursor is incorrect visible thanks to the inner borders but that would not be the case for the unstyled control (i.e. what users currently get) in the left column.
     It turns out that `div` is unnecessary, granted the money-specific code is moved to the `label`.
 - Adding CSS `class` to inputs causes them to appear on both the `label` and on the `input`. That is what creates the 2 borders in the screenshot.
     By not merging the `class` attribute on the `input`, this effect disappears and the inputs can be styled with `!important` classes.
     
This is the same code after the changes are applied to `Input`.

image

------

```
<div class="grid grid-cols-[1fr_2fr_2fr_2fr] gap-2 items-baseline">
    <script type="text/javascript" src="https://cdn.jsdelivr.net/gh/robsontenorio/mary@0.44.2/libs/currency/currency.js"></script>

    <span></span>
    <span>Default</span>
    <span>Restyled / empty</span>
    <span>Restyled / filled</span>
    <span>Money, disabled</span>
    <x-input
        label="Money" wire:model="moneyValue" prefix="USD" money />
    <x-input disabled
        class="!bg-base-100 !border-1 !border-(--input-color) !text-base-content"
        label="Money" wire:model="moneyValueEmpty" prefix="USD" money />
    <x-input disabled
        class="!bg-base-100 !border-1 !border-(--input-color) !text-base-content"
        label="Money" wire:model="moneyValue" prefix="USD" money />
    
    <span>Money, read-only</span>
    <x-input readonly
        label="Money" wire:model="moneyValue" prefix="USD" money />
    <x-input readonly
        class="!bg-base-100 !border-solid !text-base-content"
        label="Money" wire:model="moneyValueEmpty" prefix="USD" money />
    <x-input readonly
        class="!bg-base-100 !border-solid !text-base-content"
        label="Money" wire:model="moneyValue" prefix="USD" money />

    <span>Text, disabled</span>
    <x-input disabled
        label="Input"
        placeholder="Input placeholder"
        value="Input value"
    />
    <x-input disabled
        class="!bg-base-100 !border-1 !border-(--input-color) !text-base-content"
        label="Input"
        placeholder="Input placeholder"
    />
    <x-input disabled
        class="!bg-base-100 !border-1 !border-(--input-color) !text-base-content"
        label="Input"
        placeholder="Input placeholder"
        value="Input value"
    />
    <span>Text, read-only</span>
    <x-input readonly
        label="Input"
        placeholder="Input placeholder"
        value="Input value"
    />
    <x-input readonly
        class="!bg-base-100 !border-solid !text-base-content"
        label="Input"
        placeholder="Input placeholder"
    />
    <x-input readonly
        class="!bg-base-100 !border-solid !text-base-content"
        label="Input"
        placeholder="Input placeholder"
        value="Input value"
    />  
</div>
```

### DIFF
--- a/src/View/Components/DatePicker.php
+++ b/src/View/Components/DatePicker.php
@@ -157,7 +157,7 @@ class DatePicker extends Component
                                     x-data="{instance: undefined}"
                                     x-init="instance = flatpickr($refs.input, {{ $setup() }});"
                                     @if(isset($config["mode"]) && $config["mode"] == "range" && $attributes->get('live'))
-                                        @change="const value = $event.target.value; if(value.split('to').length == 2) { $wire.set('{{ $modelName() }}', value) };"
+                                        @change="const value = $event.target.value; if(value.split(instance.l10n.rangeSeparator).length == 2) { $wire.set('{{ $modelName() }}', value) };"
                                     @endif
                                     x-on:livewire:navigating.window="instance.destroy();"
                                     class="w-full"
--- a/src/View/Components/Icon.php
+++ b/src/View/Components/Icon.php
@@ -39,11 +39,13 @@ class Icon extends Component
                 @if(strlen($label ?? '') > 0)
                     <div class="inline-flex items-center gap-1">
                 @endif
-                    <x-svg :name="$icon()" {{ $attributes->class(['inline', 'w-5 h-5' => !Str::contains($attributes->get('class') ?? '', ['w-', 'h-']) ]) }}
+                    <x-svg
+                        :name="$icon()"
+                        {{ $attributes->class(['inline', 'w-5 h-5' => !Str::contains($attributes->get('class') ?? '', ['w-', 'h-']) ]) }}
                     />
 
                 @if(strlen($label ?? '') > 0)
-                        <div class="{{ $labelClasses() }}">
+                        <div class="{{ $labelClasses() }}" {{ $attributes->whereStartsWith('@') }}>
                             {{ $label }}
                         </div>
                     </div>
--- a/src/View/Components/ImageLibrary.php
+++ b/src/View/Components/ImageLibrary.php
@@ -183,8 +183,8 @@ class ImageLibrary extends Component
 
                                     {{-- ACTIONS --}}
                                     <div class="absolute flex flex-col gap-2 top-3 start-3 cursor-pointer  p-2 rounded-lg ignore-drag">
-                                        <x-mary-button @click="removeMedia('{{ $image['uuid'] }}', '{{ $image['url'] }}')"  icon="o-x-circle" :tooltip="$removeText"  class="btn-sm btn-ghost btn-circle" />
-                                        <x-mary-button @click="crop('image-{{ $modelName().'.'.$key  }}-{{ $uuid }}')" icon="o-scissors" :tooltip="$cropText"  class="btn-sm btn-ghost btn-circle" />
+                                        <x-mary-button @click="removeMedia('{{ $image['uuid'] }}', '{{ $image['url'] }}')" @touchend.prevent="removeMedia('{{ $image['uuid'] }}', '{{ $image['url'] }}')" icon="o-x-circle" :tooltip="$removeText"  class="btn-sm btn-ghost btn-circle" />
+                                        <x-mary-button @click="crop('image-{{ $modelName().'.'.$key  }}-{{ $uuid }}')" @touchend.prevent="crop('image-{{ $modelName().'.'.$key }}-{{ $uuid }}')" icon="o-scissors" :tooltip="$cropText"  class="btn-sm btn-ghost btn-circle" />
                                     </div>
                                 </div>
                             @endforeach
--- a/src/View/Components/Pin.php
+++ b/src/View/Components/Pin.php
@@ -17,6 +17,12 @@ class Pin extends Component
         public ?bool $hide = false,
         public ?string $hideType = "disc",
 
+        // Validations
+        public ?string $errorField = null,
+        public ?string $errorClass = 'text-error text-xs pt-2',
+        public ?bool $omitError = false,
+        public ?bool $firstErrorOnly = false,
+
     ) {
         $this->uuid = "mary" . md5(serialize($this)) . $id;
     }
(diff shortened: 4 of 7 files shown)


robsontenorio:
@AtmoFX  I can't reproduce it

image


The "issue" here maybe is because  the input/money component won't place `class` where you expect because of daisyUI "label + input" structure. Just use `disabled` or `readonly` without any extra classes and you are good to go.

AtmoFX:
Hey @robsontenorio,

I would like to double down on my analysis though.
If by "*the input/money component won't place `class` where you expect*", you mean that the `class` won't be applied where I **want** it to, then you are right.

TBH, I don't even see an attempt at reproducing the issues from the PR in your screenshot, since you have not copied my code nor tried to add any `class` to the inputs you copied from your help page.

Maybe you could try to do that and let me know?
Just to be sure, I don't see a `.\!border-solid` class inside the stylesheet of https://mary-ui.com/ (the other classes may be missing too), so you'll need to rebuild it for that example.

-----

If you prefer going through the long technical explanation, here it is:

With the snippet you posted, most likely, only 1 of the issues I mentioned will be visible and it won't be so obvious if you don't know what to look for. To see it:

 1. Take the disabled money field; change **nothing** in its code:
    `<x-input label="Default money" disabled wire:model="money1" prefix="USD" money />`
 2. Open the page where you inserted it.
 3. Move the cursor over the field slowly. This is what you should see: 
    mary input

From what I see `disabled` is added in 2 places in the rendered HTML and when I edit your own help page to add them in the same place, then I can at least reproduce that with no other alteration (i.e. your site's stylesheet, **as it is now**, should show you this issue).

What the switch between the default cursor and the forbidden cursor reveals is that the actual input field lies **strictly** inside the "visual" input field, with margins or padding on all 4 sides: i.e. the screen area covered by the input does not match what the user will interpret as the input.

Now, let us get into the analysis of what is going on. The easiest way to understand is most likely to add a fake CSS class and see where it lands inside the rendered HTML:
```
<x-input class="where_will_this_appear" label="Default money" wire:model="money1" prefix="USD" money />
```
becomes (after removing the extra line breaks and `<!--[if BLOCK]><![endif]-->` comments):
```
<div>
    <fieldset class="fieldset py-0">
        <legend class="fieldset-legend mb-0.5">
            Default money
        </legend>
        <label class="">
            <div class="w-full">
                <label class="input w-full where_will_this_appear">
                    <span class="label">USD</span>
                            <div class="w-full" x-data="{ amount: $wire.get('money1') }" x-init="$nextTick(() => new Currency($refs.myInput, {&quot;init&quot;:true,&quot;maskOpts&quot;:{&quot;locales&quot;:&quot;en-US&quot;}}))">
                        <input id="maryc7113191edc0aedd815586b97d162450money1" placeholder=" " x-ref="myInput" :value="amount" x-on:input="$nextTick(() => $wire.set('money1', Currency.getUnmasked(), false))" x-on:blur="$nextTick(() => $wire.set('money1', Currency.getUnmasked(), false))" inputmode="numeric" type="text" class="where_will_this_appear">
                    <input type="hidden" wire:model="money1">
                        </div>
                </label>
            </div>
        </label>
    </fieldset>
</div>
```

The `where_will_this_appear` appears in the inner `<label>` as well as in the `<input>` (scroll all the way to the right).


Armed with the above knowledge, we can infer there will be 2 types of CSS properties in the context of elements that occupy different areas on screen:
 - Properties for which children of an element are "contained" into their parent:
    For example, as long as the `<input>` and its ancestor `<label>` have e.g. the same background color, nothing will look wrong.
    However, if the `<input>` and its ancestor have different values for e.g. the cursor, it will certainly look wrong like the above capture from my computer (though less obvious in the case of the cursor).
 - Properties for which children of an element may appear distinct from their parent:
   For example, unless forced, there is no guarantee the border of an element will overlap with the border of its parent (or higher-degree ancestor).
   We do **not** want the values of those properties to be the same for the `<input>` and its ancestor `<label>` because something will look wrong if they do.
   And indeed, if no `class` (default) is added to the `x-input`, then the `<input>` will show no border while the inner `<label>` will show one (+ no ancestor of the `<input>` other than the inner `<label>` have a border).


There is also the case of properties that do not matter here, such as text color or font size: since the inner `<label>` contains no text of its own, there will never be a visual discrepancy.

robsontenorio:
I am still figuring out how we can address this case, as it is really challenging to cover all the edge cases with customizations.

AtmoFX:
Yes, you'd need to be more specific if you need my help for that but I know what you mean. It's the same for me with https://github.com/robsontenorio/mary/issues/901.
I would not be surprised whatever you decide will have impact on that issue BTW, so let me know and I'll see how to address that in a similar way as you.

AtmoFX:
@robsontenorio 

If you confirm you dislike my way of keeping the `class` merge only for the inner `label`, removing it for the `input`, maybe the solution would be to apply the same logic here as for the `header` component.
What I mean is that `header` supports separate `class` and `icon-classes`; how about changing the `input` to have `class` and `input-class`/`input-classes`?

Additionally and independently from the above, do you think it would be possible to customize the way tailwind classes are merged together? What is done right now would remain as the default but we could for instance set another one in `config/mary.php`.
Exaxmple of a candidate to override the default merging: [gehrisandro/tailwind-merge-php](https://github.com/gehrisandro/tailwind-merge-php) (though apparently not compatible with Tailwind 4 but you get the idea).
That way, all the cases where I had to use `!important` classes in my code would become regular classes.

The downside to that approach is of course that every component needs to be changed accordingly. Likely not a small endeavour.

PR #928, opened 2025-06-13
Add hidden attribute to button

akhmads:
Hi,

I added the hidden key button attribute

```
<x-button label="Create Post" :hidden="auth()->user()->cannot('create posts')" />
```

Thank you

### DIFF
--- a/src/Console/Commands/MaryInstallCommand.php
+++ b/src/Console/Commands/MaryInstallCommand.php
@@ -48,7 +48,8 @@ class MaryInstallCommand extends Command
 
         $this->info("\n");
         $this->info("✅  Done!");
-        $this->info("❤️  Sponsor: https://github.com/sponsors/robsontenorio\n");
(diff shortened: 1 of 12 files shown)


reafeichtinger:
Why not just do this?
```blade
@if(auth()->user()->cannot('create posts'))
    <x-button label="Create Post" />
@endif
```

akhmads:
@reafeichtinger  yes that's right sir, i usually do it. 
Actually this is just to shorten the code just to make it look neat when there are many buttons :)

robsontenorio:
Hey @akhmads , thanks for this.

I am yet not sure about this implementation  returning an empty string.

You can hide a button like this:

```
<x-button class="hidden"/>

<x-button label="Hi!" @class(["hidden" => $someCondition]) />
```

akhmads:
```
<x-button label="Hi!" @class(["hidden" => $someCondition]) />
```
This is a pretty good idea. Thank sir

PR #931, opened 2025-06-17
Error messages on Pin

AtmoFX:
The `Pin` field currently does not support error messages from the [form validation of Livewire](https://livewire.laravel.com/docs/forms#adding-validation).

This PR adds it to MaryUI.

Here is an example, hopefully with the exact same look and feel as `Input` fields:
 - Initial state, with no error:
    image
 - With an error:
    image


robsontenorio:
Thanks @AtmoFX

PR #933, opened 2025-06-19
ImageLibrary: add @touchend support for iOS 

xjavun:
On real iOS devices (Safari/Chrome), @click does not reliably trigger when used inside custom components such as <x-mary-button>, especially within absolute-positioned elements.

This change adds `@touchend.prevent` to the relevant buttons in the ImageLibrary component to ensure full touch compatibility on iOS devices.

This does not affect behavior on other platforms and maintains backward compatibility.
Tested on iPhone (Safari and Chrome).


robsontenorio:
Thanks @xjavun !

PR #945, opened 2025-07-08
Extend x-card with body-class option.

CaseMonster:
This tweak allows users to override the {{slot}} div with their own custom classes.

By default, the body-class is null and will work like default, only providing 'grow-1'.

If body-class is specified, ~~the body div will not use 'grow-1' and only use the specified classes.~~ the body div is appended to, including 'grow-1'.

When using the x-card, i've ran into a problem where overflow-y-auto doesn't work when using the {{slot}} of the x-card.  This ended up being the fix.

Thanks for your work on MaryUI!

### DIFF
--- a/src/View/Components/Card.php
+++ b/src/View/Components/Card.php
@@ -22,7 +22,7 @@ class Card extends Component
         public mixed $menu = null,
         public mixed $actions = null,
         public mixed $figure = null,
-        public string $bodyClass = 'grow-1'
+        public ?string $bodyClass = 'null'
     ) {
         $this->uuid = "mary" . md5(serialize($this)) . $id;
     }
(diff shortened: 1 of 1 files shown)


AtmoFX:
Hey!
Regarding `grow-1`, I just wanted to give a reminder of the purpose since I see your code may cause it to be removed (unless I have read your change incorrectly).
It comes from #892 and is to guarantee the actions slot is pushed down when a card's height is forced in some way (the example I have provided used a grid layout).

I leave it up to you and @robsontenorio to decide whether to keep the code that way or to merge the classes instead.

CaseMonster:
Yeah that's a good consideration.

I believe the most consistent, user-friendly option would be to simply append instead of replace.  I don't think a user who isn't going to dive into the attributes is going to care what they are, only that the behavior is consistent.

robsontenorio:
Thanks @CaseMonster !

image

PR #946, opened 2025-07-10
Group: allow changing the color of the selected option

themegazord:
This update adds support for customizing the color of the selected option within the `Group` component.

It works seamlessly with Tailwind's background color system, including utility classes like bg-primary, bg-secondary, and any standalone background color (e.g., bg-red-500, bg-green-200, etc.).

To use it, simply pass the utility class name (without the bg- prefix) to the `checked-class` attribute.
For example: `checked-class="primary"` will apply bg-primary to the selected option.

### DIFF
--- a/src/View/Components/Group.php
+++ b/src/View/Components/Group.php
@@ -16,6 +16,7 @@ class Group extends Component
         public ?string $label = null,
         public ?string $hint = null,
         public ?string $hintClass = 'fieldset-label',
+        public ?string $checkedClass = 'neutral',
         public ?string $optionValue = 'id',
         public ?string $optionLabel = 'name',
         public Collection|array $options = new Collection(),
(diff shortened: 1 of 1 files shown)


AtmoFX:
@robsontenorio I think that is another PR where styling Mary components could be improved the same way as we are discussing in #901, #920 and #953.
The reason why I am mentioning it is because `checked-class="primary"` allowed by this PR has the same effect as adding `class="[&:checked]:!btn-primary"` in the current main branch: like the other issues/PRs, it looks having to use an `!important` class confuses people.

robsontenorio:
@themegazord Unfortunately, this approach does not work seamlessly because Tailwind does not support or compile dynamically generated classes like `btn-$class`. 

For this reason, we provide this example in the documentation to demonstrate how to override the checked class itself.

image


I am closing this for now.

PR #948, opened 2025-07-10
Password: Allow disabling tab focus on password toggle icon

themegazord:
This PR refactors the visibility toggle buttons for password fields by introducing the new showPasswordTabstop property.

✅ What was changed:
Replaced the previous inline conditional rendering with a more flexible and accessible logic.

Introduced a new prop: showPasswordTabstop (default: false).

When showPasswordTabstop is true, the visibility toggle button receives tabindex="1" to improve keyboard accessibility.

Ensured the logic works consistently whether the icon is on the left or right, and whether placeToggleRight is used or not.

🎯 Why:
This change improves accessibility and gives developers more control over whether the visibility toggle is reachable via keyboard navigation, which is particularly useful in accessibility-focused applications.

### DIFF
--- a/src/View/Components/Password.php
+++ b/src/View/Components/Password.php
@@ -26,6 +26,7 @@ class Password extends Component
         // Password
         public ?string $passwordIcon = 'o-eye-slash',
         public ?string $passwordVisibleIcon = 'o-eye',
+        public ?bool $showPasswordTabstop = false,
         public ?bool $right = false,
         public ?bool $onlyPassword = false,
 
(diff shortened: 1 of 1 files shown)


AtmoFX:
This is a nice idea, I was working on that very same functionality myself but I was kind of temporarily stuck by related issues I found when working on it.
 1. Password fields suffer the same restyling issue as what is described in https://github.com/robsontenorio/mary/pull/920
 2. I found several things needed improvement and I was working on them all at the same time.
 See https://github.com/robsontenorio/mary/issues/950 (I had planned to open that discussion after making sure I listed every improvement I could think of but your PR kind of forced my hand 🥲)
 
 @robsontenorio: Could we perhaps have a discussion there about the few options we have to improve `password` fields?

-------

If you allow me to comment your code, the `tabindex` addition that you made twice (since the code for the toggle appears twice in the original code):
```
@if($showPasswordTabstop)
    <x-mary-button tabindex="-1" x-on:click="hidden = !hidden" class="btn-ghost btn-xs btn-circle -m-1">
        <x-mary-icon name="{{ $passwordIcon }}" x-show="hidden" class="w-4 h-4 opacity-40" />
        <x-mary-icon name="{{ $passwordVisibleIcon }}" x-show="!hidden" x-cloak class="w-4 h-4 opacity-40" />
    </x-mary-button>
@else
    <x-mary-button x-on:click="hidden = !hidden" class="btn-ghost btn-xs btn-circle -m-1">
        <x-mary-icon name="{{ $passwordIcon }}" x-show="hidden" class="w-4 h-4 opacity-40" />
        <x-mary-icon name="{{ $passwordVisibleIcon }}" x-show="!hidden" x-cloak class="w-4 h-4 opacity-40" />
    </x-mary-button>
@endif
  ```

could be simply done by adding `:tabindex="$noTabStop ? -1 : null"`.

themegazord:
You're correctly, @AtmoFX .

I don't remeber why I make like this, but, in the cleanest way it hadn't worked. If you want to leadering this issue, no problem, otherwise, if you need some help for another question, i'm available.

AtmoFX:
Hey,

If you tested #953 and are happy with it, should you close this PR?

Issue #882, opened 2025-04-24
Input Fields Accessibility


Cadranel-Kael:
Hi! I’m really loving MaryUI, it’s awesome to have an open-source alternative to FluxUI.
I noticed that the input fields are wrapped in `<fieldset>` elements. According to the W3C spec, “The `<fieldset>` element represents a set of form controls optionally grouped under a common name.” ([W3C Spec](https://www.w3.org/TR/2011/WD-html5-author-20110809/the-fieldset-element.html)).
I was wondering: why use a `<fieldset>` instead of a `<div>` or `<label>`? Also, `<legend>` elements don’t have the built-in functionality to focus the input when clicked, unlike `<label>` elements.
Furthermore, when using `<x-input>`, I noticed that there are two `<label>` elements surrounding the input.
Thanks again for your amazing work on MaryUI!

robsontenorio:
Yes , because I have mirrored daisyUI examples.

brun0f0nseca:
Hi, I also have the same problem. I have tried everything to fix it and cannot seem to "escape" this error. I have tried changing the `aria-label` to fix it, but it is impossible. I keep getting the same warning about it...

`<x-mary-select aria-label="{{ __('public.travel-simulator.from') }}" label="{{__('public.travel-simulator.from')}}" wire:model="from" :options="$itineraryStopsName" placeholder="{{__('public.travel-simulator.select-from')}}" class="w-32! p-0! bg-none [&_select]:appearance-auto" required />`

<img width="985" height="512" alt="Image" src="https://github.com/user-attachments/assets/97c3fe5e-f1ba-422f-941e-ef0ebd8ecaaf" />

PR #953, opened 2025-07-13
Password improvements

### LINKED ISSUE #942: Remove tab stop from password field
Most keyboard users use the TAB key to jump from input field to input field.

The password field has the annoying thing of stopping on the eye icon first (if the icon is on the default left side).
This is not a desired behaviour, it should skip that and go straight to the input part of the field, just like it does when adding an icon to any other field.

Other than that, love this set of components.

### LINKED ISSUE #950: Password fields improvements
Password fields seem to need 3 improvements / corrections. There is a bit to discuss before it is implemented though.

 1. The toggle button is not disabled when the field is disabled.
 2. Per my personal experience, most forms out there have no tab stop focus on toggle button.
Password visibility toggle buttons being an extra feature of MaryUI (they do not exist in DaisyUI), I feel we could just have the option to remove the tab stop; only question is: what should the default be?
    - No tab stop by default, but an attribute can enable it.
    - Tab stop by default, but an attribute can disable it.
 3. Tab order is consistent with the visual appearance of the password input i.e.:
    - If the password visibility toggle is on the left, then it gets the focus before the input
    - If the password visibility toggle is on the right, then it gets the focus after the input
  
     I feel it is more logical to always put the focus on the input first, allow users to type something, then put the focus on the toggle. That may be only personal taste on my part though, what is your take on that?

AtmoFX:
This is the implementation of the improvements for `password` fields mentioned in #950.

This PR supercedes #948 that was created concurrently, as it contains more than just addressing the mandatory tab-stop on password fields. Kudos to @themegazord for having the same idea as me (hence an obviously good one 😆) at the same time.

Among the other changes implemented, this aligns passwords with the styling logic of #920.
As a result, while it allows styling the fields in pretty much any desired way, at least AFAICT, it has the same small downsides as inputs (e.g. having to use `!important` classes for the border of the disabled fields) and both inputs and passwords will need to be aligned to whatever behavior @robsontenorio wants in the future.

----

The code:
```
<div class="grid grid-cols-[1fr_2fr_2fr_2fr] gap-2 items-baseline">

    <span></span>
    <span>Only password</span>
    <span>Left toggle</span>
    <span>Right toggle</span>

    <span>Normal</span>
    <x-password label="Test 1" only-password />
    <x-password label="Test 2" />
    <x-password label="Test 3" right />

    <span>No tab stop on toggle</span>
    <x-password label="Test 4" no-tab-stop only-password />
    <x-password label="Test 5" no-tab-stop />
    <x-password label="Test 6" no-tab-stop right />

    <span>Restyled</span>
    <x-password class="border-info" label="Test 7" only-password />
    <x-password class="border-info" label="Test 8" />
    <x-password class="border-info" label="Test 9" right />
    
    <span>Disabled</span>
    <x-password label="Test 10" disabled only-password />
    <x-password label="Test 11" disabled/>
    <x-password label="Test 12" disabled right />

    <span>Disabled, restyled</span>
    <x-password class="!border-1 !border-info" label="Test 13" disabled only-password />
    <x-password class="!border-1 !border-info" label="Test 14" disabled/>
    <x-password class="!border-1 !border-info" label="Test 15" disabled right />

</div>
```

results in:
image

Remember to change the focus on password fields by pressing tab, as this is very much included in the PR but does not render in a simple screenshot.

--
Closes #950 
Closes #942

### DIFF
--- a/src/View/Components/Password.php
+++ b/src/View/Components/Password.php
@@ -26,6 +26,7 @@ class Password extends Component
         // Password
         public ?string $passwordIcon = 'o-eye-slash',
         public ?string $passwordVisibleIcon = 'o-eye',
+        public ?bool $passwordIconTabindex = false,
         public ?bool $right = false,
         public ?bool $onlyPassword = false,
 
@@ -127,7 +128,11 @@ class Password extends Component
                                 @if($icon)
                                     <x-mary-icon :name="$icon" class="pointer-events-none w-4 h-4 opacity-40" />
                                 @elseif($placeToggleLeft())
-                                    <x-mary-button x-on:click="hidden = !hidden" class="btn-ghost btn-xs btn-circle -m-1">
+                                    <x-mary-button
+                                        x-on:click="hidden = !hidden"
+                                        class="btn-ghost btn-xs btn-circle -m-1"
+                                        :tabindex="$passwordIconTabindex ? null : -1"
+                                    >
                                         <x-mary-icon name="{{ $passwordIcon }}" x-show="hidden" class="w-4 h-4 opacity-40" />
                                         <x-mary-icon name="{{ $passwordVisibleIcon }}" x-show="!hidden" x-cloak class="w-4 h-4 opacity-40" />
                                     </x-mary-button>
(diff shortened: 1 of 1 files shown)


robsontenorio:
Hey @AtmoFX, thanks again for all the PRs!

I am happy to accept this PR if you only introduce the `noTabStop` feature, without change the html structure.

That is why I am not willing to change the structure of the HTML for the input components as you proposed in #920 and here, at leaset for now. I will consider these changes after the Livewire v4 release, as it requires a complete review of all input components to ensure consistency , manly for #882

AtmoFX:
Mistakenly closed the PR instead of correcting an issue with it.
Reopening it now to allow for tracking the change when Livewire 4 comes along.

This also resolves the conflict created by merging #958 separately, which I hope will make things easier for you.

PR #958, opened 2025-07-23
Input: add `password-icon-tabindex`

### LINKED ISSUE #942: Remove tab stop from password field

AtmoFX:
Following discussions in #953, this PR implements the no-tab-stop parameter for password fields separately.

Closes #942


robsontenorio:
Thanks again @AtmoFX !

PR #959, opened 2025-07-26
Stat: fix overflow on small screens

MohammaddReza:
added the `truncate` utility class to the description area of the `stat` component to prevent layout overflow on small screens.
Without `truncate`, long descriptions can break the layout or overflow out of the stat box.

Screenshot 2025-07-26 143146

### DIFF
--- a/src/View/Components/Stat.php
+++ b/src/View/Components/Stat.php
@@ -47,7 +47,7 @@ class Stat extends Component
                             </div>
                         @endif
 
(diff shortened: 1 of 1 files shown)


AtmoFX:
IMHO, it would be better to change the code of Mary's stat component to align it to Daisy's stat component.

robsontenorio:
Thanks!

PR #966, opened 2025-08-18
Choices/ChoicesOffline: fix Firefox issue

### LINKED ISSUE #963: Choices: input box doesn't expand for Firefox
### maryUI version

2.4.5

### daisyUI version

5.0.43

### Livewire version

3.6.3

Firefox

### What happened?

Hi, 
Just noticed that the multiple choice component input box doesn't expand for Firefox browsers - Chrome and Edge seem fine though. 

Image

chrischase011:
Closes #963 

Hello, I have added a fix for the issue in the `Choices` & `Choices Offline` component as mentioned in #963. I used `--size` as the value for min-height instead of the plain `min-h-fit` to define the initial height. Additionally, I added `h-auto` so the height automatically adjusts based on the number of selected choices.

Examples:

**Edge** 
image

**Firefox**
image

### DIFF
--- a/src/View/Components/Choices.php
+++ b/src/View/Components/Choices.php
@@ -263,7 +263,7 @@ class Choices extends Component
 
                                         {{
                                             $attributes->whereStartsWith('class')->class([
-                                                "select w-full min-h-fit pl-2.5",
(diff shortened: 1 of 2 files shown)


robsontenorio:
Thanks!

PR #974, opened 2025-09-02
Support multiple error fields in Tags component

matheus-de-araujo:
## Description:
This PR updates the Tags component to support multiple error fields, including both the array itself and its individual items.

## Key changes:
- Replaced `errorFieldName()` with `errorFieldNames()` which returns an array of possible error fields (field + field.*).
- Updated class calculation to apply `!input-error` if any error field exists.
- Updated inline Blade to display error messages for all relevant fields while keeping support for `firstErrorOnly` and `omitError`.

## Motivation:
Previously, only the error for the main array (tags) was displayed. Now, validations on `tags.*` also show correctly, providing better user feedback.

## Example of supported validation:

```blade
<x-tags label="Tags" wire:model="tags" icon="o-home" hint="Hit enter" clearable />
```

``` php
public array $tags = ['tech', 'gaming', 'art'];

public function rules()
{
     return [
        'tags' => ['required', 'array', 'size:2'],
        'tags.*' => ['required', 'string', 'min:3', 'max:50'],
    ];
}
```

### DIFF
--- a/src/View/Components/Tags.php
+++ b/src/View/Components/Tags.php
@@ -40,9 +40,19 @@ class Tags extends Component
         return $this->attributes->whereStartsWith('wire:model')->first();
     }
 
-    public function errorFieldName(): ?string
+    public function errorFieldNames(): array
     {
-        return $this->errorField ?? $this->modelName();
+        if ($this->errorField) {
+            return [$this->errorField];
+        }
+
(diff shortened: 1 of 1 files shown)


robsontenorio:
Thank you, Matheus!

I found a simpler solution.

Issue #980, opened 2025-09-23
ListItem padding class attribute.


CaseMonster:
I needed to make the ListItem a little smaller to increase the density of some of my displays.
I added an attribute to access the avatar/body padding div classes.  Do you think this change would be helpful?

<img width="652" height="241" alt="Image" src="https://github.com/user-attachments/assets/bd304350-2e8c-493e-817a-2b6362e969a2" />

<img width="545" height="1031" alt="Image" src="https://github.com/user-attachments/assets/a73b4523-6abd-419e-85b0-18eeb1879314" />

PR #981, opened 2025-09-23
ListItem: remove padding from individual elements

CaseMonster:
PR for #980


robsontenorio:
Thanks !

I changed the logic and removed padding from individual elements and placed on root div.

```php
<x-list-item ... class="!py-1" />
```

PR #986, opened 2025-10-05
DatePicker: Fix append/prepend and add clearable & placeholder

amnesca:
Hi, append/prepend are not positioned correctly (see below) and requires `@class(["w-full", "join" => $prepend || $append])` to work properly. I have also included clearable, placeholder and a method for clearing live range value when input is selected and deleted.

Thanks.

Screenshot from 2025-10-04 11-14-26


robsontenorio:
Thanks @amnesca !

PR #987, opened 2025-10-08
Drawer: add `without-backdrop-close` attribute

CaseMonster:
Optional property $withoutBackdropClose to disable the clicking outside the drawer to close.  Also disables the mouse hover effect.


robsontenorio:
Thanks KC!

image

PR #993, opened 2025-10-19
Button : Add ``disable`` attribute

Almandeel:
## Description
This pull request introduces support for dynamic disabling of buttons without relying on the ``@if`` directive or overriding the native ``disabled`` HTML attribute.


robsontenorio:
@Almandeel, I don’t quite understand your point.

You can already use it like this:

```html
<x-button label="Hi" disabled />
```
and 

```html
<x-button label="Hi" :disabled="$someCondition" />
```

Almandeel:
@robsontenorio 
Sorry about that — the issue was caused by my configuration. The syntax :disabled="$someCondition" works correctly now after adjusting my setup.

PR #997, opened 2025-10-21
fix: render placeholder without HTML escaping

4khobta:
The placeholder attribute was being HTML-escaped when rendered via  {{ $attributes->get('placeholder') }}, causing apostrophes (')  to appear as &#039;. Switched to :placeholder binding to preserve  raw text output.

https://github.com/user-attachments/assets/13f01c82-a08c-4ffd-8879-958dc969f2ee

### DIFF
--- a/src/View/Components/Input.php
+++ b/src/View/Components/Input.php
@@ -136,7 +136,7 @@ class Input extends Component
                                     {{-- INPUT --}}
                                     <input
                                         id="{{ $uuid }}"
-                                        placeholder="{{ $attributes->get('placeholder') }} "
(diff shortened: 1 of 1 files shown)


robsontenorio:
This works well without any changes.

Single quote and backtick: 

```html
<x-input label="Inline 'a `a" wire:model="name" placeholder="Inline 'a `a" inline />
```

What string are you trying?

4khobta:
Yes, I overdid it)

String:
`:placeholder="__('Name')"  // "Name": "Ім'я"`
...
I replace with:
`placeholder="{!!__('Name')!!}" `

thanks

Issue #976, opened 2025-09-03
Suggestion: Add progress bar indicator to toasts before auto-dismiss


lucas-davi:
Add a new option to display a progress bar inside toasts, indicating the remaining time before they auto-dismiss. This will provide users with a clear visual cue of how long the tooltip will stay visible.

For example:

<img width="482" height="140" alt="Image" src="https://github.com/user-attachments/assets/9aace8ec-5cc9-45e9-82fb-d11d3cdc5366" />

Thanks for support.

lucas-davi:
After updating the package, I tested the toast with a progress indicator. However, the toast is disappearing before the progress bar finishes. It seems the toast is being closed automatically earlier than expected.

![Image](https://github.com/user-attachments/assets/5f72576b-6ce2-4549-89cc-3272f48decc3)

robsontenorio:
cc @lchevalot 

Here it works fine. Maybe clear the cache?

```
php artisan view:clear
```

lucas-davi:
After clearing the cache, the behavior remains the same.

lchevalot:
@lucas-davi can you provide the code sample of your toast ? Along with checking the console for any warning/error please ?

also note that it may be due to the progress bar being a bit jerky on local server development, that happened to me when my RAM was almost fully used (i can't really figure out why), let me know if that's your case

lucas-davi:
I checked and there no erros in the console.

my code

```blade
<x-mary-button icon="o-chat-bubble-bottom-center-text" label="Test Toast" wire:click="testToast" />
```

```php
public function testToast()
{
    $this->success('This is a success toast message.');
}
```

I actually tested it in my local development environment, but the memory usage seems to be relatively fine.

lchevalot:
Then this is probably due to the Vite/Alpine local environment being jerky (imprecise setInterval, heavy sourcemaps, ...). Try to run your code locally with `npm run build` instead of `npm run dev` and you may see the difference

edit: i can see the same issue in production after triggering a same toast multiple times ... i'll investigate when i get the time

lchevalot:
[...]

PR #1001, opened 2025-11-06
Toast: Add progress support

lchevalot:
This is the improvement mentionned in #976.

This PR adds a progress bar under the Toast component by default. The progress bar will show up with the start value of **100**, and decrease until **0** according to the duration of the toast, just like this :

image

The progress bar can be removed by setting the `noProgress` variable to true like this : `$this->*toastType*(noProgress: true)`

Moreover, i added a feature that allows the timeout (along with the progress bar) to stop when hovering the toast, and resume when leaving the hover.

Since the progress bar is the `<progress>` component from **DaisyUI**, we can use the [progress classes](https://daisyui.com/components/progress/) to customize the it with the `progressClass` variable like this : `$this->*toastType*(progressClass: 'progress-primary')`

Note that `progressClass` is set to `null` by default, so the default color of the progress bar will be **neutral**.

### DIFF
--- a/src/Exceptions/ToastException.php
+++ b/src/Exceptions/ToastException.php
@@ -25,7 +25,7 @@ class ToastException extends Exception
 
     protected bool $preventDefault = true;
 
-    protected bool $progress = false;
+    protected bool $noProgress = false;
 
     protected ?string $progressClass = null;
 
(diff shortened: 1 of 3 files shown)


robsontenorio:
@lchevalot 

Yes, I agree you can change the default to `progress = true`. But you have to invert the logic across the code to `noProgress = false`.

Another missing cool feature is the ability to pause the progress if you mouse hover it. Could you please give a shot?

After that, is all good to merge.

Thanks you!

lchevalot:
@robsontenorio thank you for your feedback ! i'll get on that soon :)

lchevalot:
@robsontenorio All good now !

robsontenorio:
Thanks, Lucas!

PR #1004, opened 2025-11-07
Choices: make searchable input react to Cyrillic by switching from keydown to input debounce

4khobta:
Searchable field didn't trigger search() on Cyrillic input because keydown events were not firing properly.
Replaced @keydown.debounce with @input.debounce to fix it.

### DIFF
--- a/src/View/Components/Choices.php
+++ b/src/View/Components/Choices.php
@@ -333,7 +333,7 @@ class Choices extends Component
                                                  @endif
 
                                                 @if($searchable)
-                                                    @keydown.debounce.{{ $debounce }}="search($el.value, $event)"
+                                                    @input.debounce.{{ $debounce }}="search($el.value, $event)"
                                                 @endif
                                              />
                                         </div>


robsontenorio:
@4khobta 

What do you mean by "Cyrillic"?
Can you give me some examples on how it breaks?

4khobta:
Does not respond to the Ukrainian alphabet, for example
[Screen record](https://github.com/user-attachments/assets/62bb657f-d447-4bd0-9cd1-0bb0714ad666)

robsontenorio:
I don't have this kind of keyboard.

But, the virtual keyboard works fine on online docs.
https://mary-ui.com/docs/components/choices#searchable-server

Maybe someone that use an special keyboard can give a hint in here.


https://github.com/user-attachments/assets/c46e0db4-e673-4816-a4ce-a5a097c048b3

4khobta:
Checked. Doesn’t work on Ubuntu 24.04 for Google Chrome 142.0.7444.162 and Chromium 142.0.7444.59. Works in Opera and Firefox, and also works in Google Chrome on Android and Windows. Maybe someone else can test it as well. Thanks

ShotaFujimoto:
Check the environment
- MacOS Version 15.3.2 (24D81)
- Google Chrome Version 142.0.7444.135 (Official Build) (arm64)
- Input format "Japanese" (JIS keybord)

I checked this environment and it worked without any problems.

4khobta:
Checked. Chrome/Chromium on Wayland does not fire `keydown` events for Ukrainian (and other non-Latin) layouts. 
`keypress` and `keyup` still fire, but `keydown` is missing entirely.

The same code works correctly on:
– Chrome/Chromium with `--ozone-platform=x11`
– Firefox / Opera on Wayland
– Chrome on Windows / Android

So this is a Chromium + Wayland keyboard event bug, not an issue in the component.
Bug report registered.

Thanks

PR #1006, opened 2025-11-10
MenuItem: improve support for using route parameter

AlwynW:
This modification allows for use with the (named) route parameter only without having to set the link explicitly. 
Previously if you didn't set a link, the route wouldn't activate if on that page. Also the route parameter was ignored when constructing the link.

There's now also support for route parameters.

```
<x-menu-item title="Dashboard" route="dashboard" /> // Or with parameters
<x-menu-item title="User Profile" route="users.show" :route-params="['id' => 1]" />
```

### DIFF
--- a/src/View/Components/MenuItem.php
+++ b/src/View/Components/MenuItem.php
@@ -19,6 +19,7 @@ class MenuItem extends Component
         public ?string $spinner = null,
         public ?string $link = null,
         public ?string $route = null,
+        public ?array $routeParams = [],
         public ?bool $external = false,
         public ?bool $noWireNavigate = false,
         public ?string $badge = null,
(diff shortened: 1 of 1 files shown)


robsontenorio:
@AlwynW 

Can you please give me some usage examples?

AlwynW:
Sure.
I always prefer to use named routes when I set up navigation. Just in case if the url changes it doesn't break navigation.

Example from my current project where I'm using MaryUI (Volt but would work the same with Route):

```
Volt::route('/memberships', 'user.memberships')->name('memberships');
Volt::route('/messages/{folder}', 'user.messages')->name('messages');

```
I can link them like this:

```
<x-menu-item title="Memberships" icon="m-bookmark-square" route="memberships" />
<x-menu-item title="Inbox" icon="m-inbox" route="messages" :route-params="['folder' => 'inbox']" />
```

robsontenorio:
You missed the html part on how you would use it.

AlwynW:
Weird, github always seems to cut off the last sentences 😅. Fixed reply.

robsontenorio:
Thanks, Alwyn!

PR #1007, opened 2025-11-12
Header: add `use-h1` attribute

### LINKED ISSUE #1002: Suggestion: make <x-header> use <h1> tags
Good day,

First, I'm new to all this Github interaction. 
I apologize if it's not the way to do things.

I was implementing some SEO on a project using MaryUI and I discovered that the <x-header> does not use the `<h1></h1>` tag.
However, good SEO practices require one unique `<h1></h1>` title per page as I understand it.

Would it be possible to add an attribute to the <x-header> component such as "h1" so that the component will specifically use the `<h1></h1>` HTML tag when rendering?

Thank you for your consideration,
Joey

josalzi:
Good day,

Supporting #1002, I have modified the Header component view file to include an optional use-h1 attribute.
If the attribute is passed to the component, the render will be done using `<h1></h1>` tags.
This is to support SEO optimization. 

Usage: `<x-header title="Welcome!" size="text-4xl" subtitle="This is a welcome page" class="!mb-5" use-h1 />`

From my test, there is no breaking of existing styling/layout, and you can still override default h1 styling references with custom class attributes.

The html rendering looks correct as well: 
`<div id="welcome" class="mb-10 !mb-5"><div class="flex flex-wrap gap-5 justify-between items-center"><div><h1 class="flex items-center text-4xl font-extrabold " >Welcome!</h1><div class="text-base-content/50 text-sm mt-1 " >This is a welcome page</div></div><div class="flex items-center gap-3 " > </div></div>`

Let me know your thoughts.
Joey

Note: sorry for the bad readability of my PR, I have difficulties with the Markdown editor.
Note 2: following advice, new PR to commit from a branch, and not from the main branch on my fork.

Closes #1002

### DIFF
--- a/src/View/Components/Header.php
+++ b/src/View/Components/Header.php
@@ -11,6 +11,8 @@ class Header extends Component
 {
     public string $anchor = '';
 
+    public string $titleTag = 'div';
+
     public function __construct(
         public ?string $title = null,
         public ?string $subtitle = null,
@@ -30,6 +32,7 @@ class Header extends Component
         public mixed $actions = null,
     ) {
         $this->anchor = Str::slug($title);
+        $this->titleTag = $useH1 ? 'h1' : 'div';
     }
 
     public function progressTarget(): ?string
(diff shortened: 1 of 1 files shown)


josalzi:
Good day,

Is there anything else I can do to get this merged? 
Or how can I improve my request? 

Thank you in advance.

robsontenorio:
Instead of duplicating the code, could you simplify it by switching the tag itself (div/h1)?

Besides that, it must preserve all examples mentioned in the documentation.
https://mary-ui.com/docs/components/header

josalzi:
Hello,

Thank you for pointing me toward the right direction.
I have simplified the code and prevented inelegant code duplication.

Regarding compliance with existing documentation, I do not see why any features would be lost to be honest.
So, for me, any example in the documentation will work out of the box, regardless of the use of "use-h1" attribute. 

Tested locally, nothing wrong seems to happen.
Let me know if I can help you furthermore with my request.

robsontenorio:
Thanks @josalzi !


image

PR #1008, opened 2025-11-12
Added autocomplete attribute to Input/Password components

josalzi:
Good day,

Autocomplete attribute has become standards when designing web forms.
Especially, when using MaryUI components to design a simple log-in form, Chrome Dev Console will throw: connexion:1 [DOM] Input elements should have autocomplete attributes (suggested: "current-password"): (More info: https://goo.gl/9p2vKq) <input id=​"marye6ee36000329b1f79d37240464ded95cpassword" placeholder=​" " x-bind:type=​"hidden ? 'password' :​ 'text'" wire:model=​"password" type=​"password">​flex

Added feature to pass on an optional autocomplete attribute and value to the following components:
- Input
- Password

### DIFF
--- a/src/View/Components/Input.php
+++ b/src/View/Components/Input.php
@@ -23,6 +23,7 @@ class Input extends Component
         public ?bool $clearable = false,
         public ?bool $money = false,
         public ?string $locale = 'en-US',
+        public ?string $autocomplete = null,
 
         // Slots
         public mixed $prepend = null,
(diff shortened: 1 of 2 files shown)


robsontenorio:
@josalzi 

There is no need to implement this.
Because any attribute is already automatically injected.

```html
<x-input label="E-mail" anything="something"  autocomplete="email" />
```

PR #1017, opened 2025-12-08
Update choices and choices offline update highlight and keyboard navigation concept

akhmads:
image

Hello,

I'm trying to modify the hover and keyboard navigation behavior of options, so that highlighting the active option no longer uses focus or tabindex, but instead uses the active index of that option programmatically.

The point is that when we move focus to select an option, we don't lose focus on the keyword input box cursor.

I'm just concerned about a bug, so can this code be reviewed to see if it's applicable?

Thank you.

### DIFF
--- a/src/Exceptions/ToastException.php
+++ b/src/Exceptions/ToastException.php
@@ -25,14 +25,20 @@ class ToastException extends Exception
 
     protected bool $preventDefault = true;
 
+    protected bool $noProgress = false;
+
+    protected ?string $progressClass = null;
+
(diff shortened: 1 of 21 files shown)


robsontenorio:
This component is complex and dangerous enough to require this much logic and code.

We already have keyboard navigation with arrow up and down.

Check out this example:
https://mary-ui.com/docs/components/choices#searchable-server


Thanks for this, but unfortunately I am not merging this for now.

akhmads:
It's okay, thank you.

PR #1021, opened 2025-12-11
Header: Add actions check before adding the container

### LINKED ISSUE #1020: Header: actions div is causing a gap on modals title
### maryUI version

2.4.9

### daisyUI version

5.5.11

### Livewire version

3.6.4

Chrome

### What happened?

Currently, the header component allows :

1. A title, with a subtitle and icon if wanted
2. A middle element
3. Actions elements

However, even if we don't want actions elements, the `<div>` containing them will still show up because we don't make a check on actions existence like middle elements existence, hence the **Header.php** code sample below :

```
// here we check if middle exists before putting the div
@if($middle)
    <div @class(["flex items-center justify-center gap-3 grow order-last sm:order-none", is_string($middle) ? '' :$middle?->attributes->get('class')])>
         <div class="w-full lg:w-auto">
              {{ $middle }}
          </div>
     </div>
@endif

// here we put actions div regardless if there are actions or not
<div @class(["flex items-center gap-3", is_string($actions) ? '' : $actions?->attributes->get('class') ]) >
    {{ $actions}}
</div>
```

this cause the title to have a useless gap on modals when the flex-wrap occurs (i didn't find that anywhere else), like this on chrome :

With empty actions `<div>`
Image

Without empty actions `<div>`
Image

The same thing happens on firefox (and probably other browsers too).

This is just small details, but i'm going to do a PR for this small fix :)

lchevalot:
Fix of the issue mentionned in #1020.

This commit just checks if there are actions before actually adding the `<div>` that contains them. Doing so removes the useless gap in the header when there are no actions.

Closes #1020

### DIFF
--- a/src/View/Components/Header.php
+++ b/src/View/Components/Header.php
@@ -11,6 +11,8 @@ class Header extends Component
 {
     public string $anchor = '';
 
+    public string $titleTag = 'div';
+
     public function __construct(
         public ?string $title = null,
         public ?string $subtitle = null,
@@ -19,6 +21,7 @@ class Header extends Component
         public string $progressIndicatorClass = "progress-primary",
         public ?bool $withAnchor = false,
         public ?string $size = 'text-2xl',
+        public ?bool $useH1 = false,
 
         // Icon
         public ?string $icon = null,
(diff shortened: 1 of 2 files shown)


robsontenorio:
Thanks, Lucas!

PR #1026, opened 2025-12-23
Option to Disable Routes

alloylab:
provides the ability within the config to disable routes; reasoning behind this is if you aren't using them you should have the ability to disable them rather than leaving the route available, also this commit fixes the typo, toogle ... it should be toggle


robsontenorio:
I got your idea. But, your PR breaks the things if you set enabled to `false`

By the way, if you do not use the routes, just leave them.
Or just override them with a 403 or something on your app.


image

PR #1031, opened 2025-12-30
DateTime: Fix missing "join" class when there are prepend/append elements

### LINKED ISSUE #1025: Date Time's append and prepend slots render on separate lines.
### maryUI version

2.4.9

### daisyUI version

5.4.7

### Livewire version

3.7.0

Firefox, Chrome

### What happened?

This code...

```
<x-datetime label="Append a button">
    <x-slot:append>
        <x-button label="I am a button" class="join-item btn-primary" />
    </x-slot:append>
</x-datetime>
```

... displays like...

Image

I think the issue has to do with line 73 of DateTime.php: `<div class="w-full">`. 

Changing that to `<div @class(["w-full", "join" => $prepend || $append])>` (similar to line 111 of Input.php), seems to fix the issue...

Image

lchevalot:
Fix of the issue mentionned in #1025.

Closes #1025

### DIFF
--- a/src/View/Components/DateTime.php
+++ b/src/View/Components/DateTime.php
@@ -70,7 +70,7 @@ class DateTime extends Component
                             <span class="font-semibold">{{ $label }}</span>
                         @endif
 
-                        <div class="w-full">
+                        <div @class(["w-full", "join" => $prepend || $append])>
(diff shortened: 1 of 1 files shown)


robsontenorio:
Thanks Lucas!

PR #1034, opened 2025-12-31
Carousel: accessibility improvements

### LINKED ISSUE #1022: Suggestion: Add new field to carousel
I think it would be good to add lazy="loading" and the alt tag to the image. Because there are many sites for testing, and this appears as a performance warning.

Like:
```
<!-- Image -->
<img lazy="loading" alt="data_get($slide, 'imageAlt')" class="w-full h-full inset-0 object-cover" src="{{ data_get($slide, 'image') }}" />
```

lchevalot:
Improvements mentionned in #1022 :

Added a check on the lazy attribute of a slide. If there's a lazy attribute, the image will be set to lazy loading. If not, it will be set to eager loading (default behaviour).

Improvements mentionned in #882 :

To get along accessibility standards, i've made a few changes.

- Unlabelled buttons have now the `aria-label` attribute, which includes the previous/next buttons but also the buttons to select the slide.
- Added a check on the alt attribute to add images an alternative text.
- Increased the gap of the buttons for slide selection (was throwing a warning about accessibility because the gap was too tight).

Closes #1022

### DIFF
--- a/src/View/Components/Carousel.php
+++ b/src/View/Components/Carousel.php
@@ -138,6 +138,7 @@ class Carousel extends Component
                                 class="w-full h-full inset-0 object-cover"
                                 src="{{ data_get($slide, 'image') }}"
                                 alt="{{ data_get($slide, 'alt') }}"
+                                loading="{{ data_get($slide, 'lazy') ? 'lazy' : 'eager' }}"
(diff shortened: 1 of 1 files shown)


robsontenorio:
Thanks, Lucas!

PR #1036, opened 2026-01-02
Modal: add custom header slot

### LINKED ISSUE #901: Font issues with title of modals

lchevalot:
Mentionned in #901.

Allows the user to either set a title with the default DaisyUI style with the `title` attribute, or a custom title slot if more customization is needed.

**Breaking change :** none

Closes #901

### DIFF
--- a/src/View/Components/Modal.php
+++ b/src/View/Components/Modal.php
@@ -10,14 +10,14 @@ class Modal extends Component
 {
     public function __construct(
         public ?string $id = '',
-        public ?string $title = null,
+        public mixed $title = null,
+        public ?string $subtitle = null,
         public ?string $boxClass = null,
         public ?bool $separator = false,
         public ?bool $persistent = false,
         public ?bool $withoutTrapFocus = false,
 
         // Slots
-        public mixed $header = null,
         public ?string $actions = null
     ) {
         //
@@ -56,12 +56,17 @@ class Modal extends Component
                             </form>
                         @endif
 
-                        @if($header)
-                            <div>
-                                {{ $header }}
+                        @if(is_string($title))
+                            <div class="mb-5">
+                                <h3 class="font-bold text-lg">{{ $title }}</h3>
+                                @isset($subtitle)
+                                    <div class="text-base-content/50 text-sm mt-1">
+                                        {{ $subtitle }}
+                                    </div>
+                                @endisset
                             </div>
-                        @elseif($title)
-                            <h3 class="font-bold text-lg mb-5">{{ $title }}</h3>
+                        @else
+                            {{ $title }}
                         @endif
 
                         <div>
--- a/src/View/Components/Modal.php
+++ b/src/View/Components/Modal.php
@@ -11,13 +11,13 @@ class Modal extends Component
     public function __construct(
         public ?string $id = '',
         public ?string $title = null,
-        public ?string $subtitle = null,
         public ?string $boxClass = null,
         public ?bool $separator = false,
         public ?bool $persistent = false,
         public ?bool $withoutTrapFocus = false,
 
         // Slots
+        public mixed $header = null,
         public ?string $actions = null
     ) {
         //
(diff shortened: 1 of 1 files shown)


robsontenorio:
@lchevalot Why cant we keep the `subtitle` atribute ?

If `header` is provided, so it uses the custom slot.
If not, uses the default that is the `x-header` component.

lchevalot:
@robsontenorio if we consider the use of a **custom slot** for using the **Header component** because this wouldn't be the default behaviour anymore (the new default being just a `<h3>` tag with the title), then the subtitle attribute would become useless.

The main issue here might just be the lack of documentation in the modal page about how we could turn the
```blade
<x-modal wire:model="myModal1" title="Hey" class="backdrop-blur">
...
```
into 
```blade
<x-modal wire:model="myModal1" class="backdrop-blur">
    <x-slot:title>
        <x-header ... />
    </x-slot:title>
...
```
for more flexibility.

As mentionned by @AtmoFX and in my opinion, having the header as a default title for the modal isn't the best because the font isn't matching DaisyUIs one and it makes the act of changing the header is a bit overcomplicated, maybe having a component inside another component by default isn't the best thing, especially if we can override this by just adding slots.

Tell me your thoughts about this, you can close the draft PR if you disagree.

robsontenorio:
We could no longer use `x-header` as the default, but we should provide backward compatibility APIs for those who need all the attributes as before.

Is it possible to adjust your PR for this?

lchevalot:
I'll take a look onto that.

lchevalot:
As requested, the `$subtitle` attribute is still present.
I also made the `$title` **mixed** to allow it to be either a **string for the default title style**, or a **slot for customization**, allowing the removal of the useless `$header` slot. Here's an example :

```blade
<!-- Modal with default title (and subtitle) -->
<x-modal wire:model="myModal1" title="Title without header" subtitle="With a subtitle" class="backdrop-blur">
    Press `ESC`, click outside or click `CANCEL` to close.

    <x-slot:actions>
        <x-button label="Cancel" @click="$wire.myModal1 = false" />
    </x-slot:actions>
</x-modal>

<!-- Modal with custom title slot (here the header component) -->
<x-modal wire:model="myModal2" class="backdrop-blur">
    Press `ESC`, click outside or click `CANCEL` to close.
    <x-slot:title>
        <x-header title="Title with header" subtitle="With a subtitle" class="!mb-5"></x-header>
    </x-slot:title>

    <x-slot:actions>
        <x-button label="Cancel" @click="$wire.myModal2 = false" />     </x-slot:actions>
</x-modal>
```

Modal with default title style (and subtitle) :
image

Modal with custom title slot (header component here) :
image

If this solution suits you, the only suggestion i would give is to add a bit of documentation on the modal page to show how we can set either
```blade
<x-modal title="Title">
```
or
```blade
<x-modal>
    <x-slot:title>
        ...
    </x-slot:title>
```

robsontenorio:
Hey @lchevalot  can you DM me on Twitter or Discord before going deeper with this?

https://x.com/robsontenorio
https://discord.gg/c2Dv8T2X2s

PR #1037, opened 2026-01-02
Header: add weight attribute for font weight customization

### LINKED ISSUE #901: Font issues with title of modals

lchevalot:
Mentionned in #901.

Allows the user to change the font weight of the modal title if needed with the `weight` attribute.

Closes #901

### DIFF
--- a/src/View/Components/Header.php
+++ b/src/View/Components/Header.php
@@ -21,6 +21,7 @@ class Header extends Component
         public string $progressIndicatorClass = "progress-primary",
         public ?bool $withAnchor = false,
         public ?string $size = 'text-2xl',
+        public ?string $weight = 'font-extrabold',
         public ?bool $useH1 = false,
 
         // Icon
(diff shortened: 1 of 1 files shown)


robsontenorio:
Thanks, Lucas!

Issue #1048, opened 2026-01-29
Choices/Choices-offline: Escape Values Option


amnesca:
Hi, sometimes it's handy to create choice values on the fly without having to create/maintain separate models for every filter. This works fine by mapping the unique values to both the 'id' and 'name', however, if the value contains single or double quotes then the 'id' value is malformed in the AlpineJS code. Escaping it during mapping doesn't work due to the toggle 'getOptionValue' method. So to fix this, could we provide an 'escapeValue' conditional inside the 'getOptionValue' method like below?

Thanks.

```diff
+public ?bool $escapeValues = false

public function getOptionValue($option): mixed
{
	$value = data_get($option, $this->optionValue);

	if ($this->valuesAsString) {
		return "'$value'";
	}

+	if ($this->escapeValues) {
+   	$value = addslashes($value);
+
+     	return "'$value'";
+ 	}

	return is_numeric($value) && ! str($value)->startsWith('0') ? $value : "'$value'";
}
```

robsontenorio:
If it does not break other features, go ahead.

amnesca:
Done - refer [PR #1050](https://github.com/robsontenorio/mary/pull/1050#issue-3883867232)

PR #1050, opened 2026-02-02
Choices/Choices-offline: add escape values conditional

amnesca:
Hi, this request adds an 'escapeValues' conditional to prevent malformed ids in AlpineJS (eg. values containing single/double quotes etc.) when creating choice options on the fly by mapping unique values. Refer [Issue #1048](https://github.com/robsontenorio/mary/issues/1048#issue-3868372908).


robsontenorio:
Thanks !

PR #1055, opened 2026-02-10
Toast: fix progress bar class

amnesca:
The recent progress bar addition to the Toast component has a fixed class of  ```rounded-md```, which works fine for the default DaisyUI theme, however, other themes with different rounded classes don't match (see below). 

**theme: bumblebee**
Toast-progress-bug

**Fix:** 
Changing the main toast class from ```rounded-md``` to ```rounded-box``` makes it dynamic and matches other theme settings, as seen below:

**theme: light**
Toast-progress-fix-light

**theme: bumblebee**
Toast-progress-fix-bumblebee

### DIFF
--- a/src/View/Components/Toast.php
+++ b/src/View/Components/Toast.php
@@ -114,7 +114,7 @@ class Toast extends Component
                     @mary-toast.window="start($event.detail.toast)"
                 >
                     <div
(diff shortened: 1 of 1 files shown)


robsontenorio:
Thanks!

Issue #1057, opened 2026-02-16
Popover bugging password input


joaodemarco:
### maryUI version

2.7.0

### daisyUI version

5.4.2

### Livewire version

3.7.10

### What browsers are affected?

Firefox, Chrome, Microsoft Edge

### What happened?

I'm trying to use a **password** components and it works as it should, except when i set a popover to any value. It seems like the input starts to take the whole screen (as it shows the IBeam cursor, and when i click anywhere in the screen it focus the password input). If I use the password component without the popover, it works just fine.

HTML/Blade

```blade
            <x-mary-password
            label="Password"
            popover="the text of the popover doesn't matter"
            wire:model.blur="form.password"
            right
            />
```

**OBS:** i can't test on safari since i dont have a mac, but i dont think it's a browser issue

amnesca:
Hi joaodemarco, this is because there is no attribute for popovers like there is in the [input component](https://github.com/robsontenorio/mary/blob/main/src/View/Components/Input.php). @robsontenorio should this have been added to the password component?

robsontenorio:
Yes, `<x-password>` does not have a `popover` attribute.

PR are welcome.  Can done in the same way we did it in `<x-input>`

PR #1059, opened 2026-02-21
Password: add popover attribute

amnesca:
Add popover to password component props.

Closes [#1057](https://github.com/robsontenorio/mary/issues/1057)


robsontenorio:
Thanks!

PR #1073, opened 2026-03-17
All input components: add `popover` option

slam357:
Adds popover support to all form components (previously only available for input elements);
Adds support for setting custom classes for popovers;
Usage:
`<x-input popover popover-trigger-class="opacity-50" popover-content-class="max-w-50" />`


robsontenorio:
Thanks!

Issue #1072, opened 2026-03-15
x-table setting sort-by ignores different name


rambousek:
### maryUI version

2.8.0

### daisyUI version

5.0.9

### Livewire version

4.2.1

### What browsers are affected?

Chrome

### What happened?

I'm using several <x-table> in one blade and I want to have sort-by for each of them.

I have following setup

```php
    public array $sortByEmployee = ['column' => 'surname', 'direction' => 'asc'];

.....

            ->orderBy(...array_values($this->sortByEmployee))

```

```blade
 <x-table :headers="$headersEmployee" :rows="$employees" :sort-by="$sortByEmployee" wire:model="sortByEmployee">
```

When I first open the page, table is sorted by correct column based on default $sortByEmployee.
When I click on any column, I get the following error:
_Unable to set component data. Public property [$sortBy] not found on component_

robsontenorio:
Actually this is a restriction. Because the `x-table` expects to update a variable exactly named as  `sortBy`.
PR are welcome.

```php
public string $sortByProperty = 'sortBy',
```

https://github.com/robsontenorio/mary/blob/ca5303acc3e3f6d2c8b042739d6b9fbbdc85e8f5/src/View/Components/Table.php#L312-L314

Almandeel:
#1079

robsontenorio:
Solved

PR #1079, opened 2026-04-03
Table: add alternative `sort-by-property`  in `table`

Almandeel:


robsontenorio:
@Almandeel  Could you please check again my comment? https://github.com/robsontenorio/mary/issues/1072#issuecomment-4112623989

You have added a new set of variables to handle sorting, along with its logic.

However, the simplest approach is to simply define which property should be "bound" and change the correspondent line of code. So you can use all the existing logic.

Almandeel:
@robsontenorio Updated

robsontenorio:
Thanks @Almandeel !


---

image

PR #1087, opened 2026-05-08
Fix: use logical margin utility for RTL support

mrezdev:
This PR improves RTL/LTR layout compatibility by replacing Tailwind's physical left margin utility `-ml-1` with the logical start margin utility `-ms-1`.

## Changed Components

The following components were updated:

- Colorpicker
- DatePicker
- DateTime
- Select
- SelectGroup

`-ml-1` always applies a negative margin to the left side, which works correctly in LTR layouts but can create incorrect spacing in RTL layouts.

Using `-ms-1` applies the negative margin to the inline-start side:

- In LTR: it behaves like `-ml-1`
- In RTL: it behaves like `-mr-1`

This makes the components direction-aware and improves RTL support without changing the visual behavior for existing LTR layouts.


robsontenorio:
Thanks!

PR #1113, opened 2026-08-12
Pin: sync with external changes

behvandi11:
### Bug
When the value bound to `wire:model` on `<x-mary-pin>` is changed
from the server side (e.g. `$this->code = ''` inside a Livewire
component action), the visible pin boxes do not update. This is
because each box is bound to a local Alpine `inputs[i]` array, and
`value` (the entangled property) is only ever written *from*
`inputs` via `handlePin()` — never the other way around. There was
no listener syncing external changes to `value` back into `inputs`.

### Reproduction
1. Render `<x-mary-pin wire:model="code" size="6" />`.
2. Type a code into the boxes.
3. From a Livewire action, set `$this->code = ''` (e.g. on a
   "resend code" button) and re-render.
4. Expected: boxes are cleared. Actual: boxes still show the old
   digits, even though the underlying model is now empty.

### Fix
Added a `$watch('value', ...)` inside `init()` that re-syncs
`inputs` from `value` whenever it changes — guarded so that
changes originating from `handlePin()` itself (typing/pasting) are
ignored, avoiding redundant writes and any risk of a feedback loop.

### Compatibility
- No changes to markup, props, or public API.
- No behavior change for typing, backspace, or paste — verified
  manually.
- Negligible overhead: one array `join` + string comparison per
  actual value change.


robsontenorio:
Thanks @behvandi11

PR #1114, opened 2026-08-20
Dropdown: add popover to escape clipping ancestors

lauroguedes:
# Dropdown: add `popover` to escape clipping ancestors

## Bug

A `<x-mary-dropdown>` inside `<x-mary-main>`'s collapsible sidebar gets clipped by the sidebar instead of floating over the page. It is easiest to see once the sidebar is collapsed to its 62px rail.

The sidebar wrapper from `Main.php` has `overflow-x-hidden overflow-y-auto`, and daisyUI adds `translate: 0%` to that same element. Each one traps the panel on its own:

* `overflow-x-hidden` clips it. We cannot relax that, because the spec says one axis cannot be `visible` while the other is `auto`, and dropping `overflow-y` would kill the sidebar scrolling.
* A `translate` other than `none` turns the element into a containing block, and a containing block ancestor clips `position: fixed` children too, so switching the anchor strategy does not help either.

The only thing that escapes both is the top layer, so this uses the Popover API.

Screenshot 2026-08-20 at 8 58 59 PM

## Reproduction

1. Any app using `<x-mary-main>` with `<x-slot:sidebar collapsible>`.
2. Put an `<x-mary-dropdown>` inside the sidebar.
3. Collapse the sidebar.
4. Open the dropdown. The panel is clipped to the rail.

## Fix

A new opt-in `popover` prop. Nothing changes unless you pass it.

When it is on, the panel gets the `popover` attribute so the browser promotes it to the top layer, and the anchor switches to Floating UI's `fixed` strategy, which is what a top-layer element needs. Alpine's existing `open` state still drives it through `showPopover()` and `hidePopover()`, and a `@toggle` listener syncs that state back when the browser light-dismisses the panel on Esc or an outside click. The component API stays the same.

Three details in the diff look like noise but are not:

* **`inset-auto m-0`**: the browser's own stylesheet gives every popover `inset: 0` and `margin: auto`. Alpine's anchor only writes `left` and `top`, so the leftover `right` and `bottom` stretch the panel from its anchor all the way to the right edge of the screen.
* **`[&[popover]_.mary-hideable]:!block`**: a collapsed sidebar hides labels with `[&_.mary-hideable]:!hidden` on its wrapper. The top layer changes how the panel is painted, not where it sits in the DOM, so it is still a descendant and its labels still match that rule. The `[popover]` part is required. Without it both selectors land on the same specificity with `!important` on each side, and Tailwind writes `block` before `hidden`, so the label stays hidden. Adding the attribute raises the specificity and the result stops depending on output order.
* **The guard in the constructor**: `popover` and `no-x-anchor` cannot work together. `no-x-anchor` hands positioning to daisyUI, which offsets `dropdown-content` against the `.dropdown` ancestor. In the top layer the viewport becomes the containing block, so those offsets resolve against the screen and the panel lands in the top-left corner, ignoring `top` and `right`. Throwing at construction matches how `Choices` and `Password` already reject bad prop combinations.

### Usage

```blade
<x-mary-dropdown right popover>
    <x-slot:trigger>
        <x-mary-avatar :image="$user->avatar" />
    </x-slot:trigger>

    <x-mary-menu-item title="Settings" icon="s-cog-6-tooth" link="/settings" />
</x-mary-dropdown>
```

Screenshot 2026-08-20 at 9 00 10 PM

## Compatibility

* Opt-in, so nothing changes for anyone who does not pass the prop.
* The new parameter is last, so positional arguments still line up.
* Popover API is Baseline: Chrome 114+, Edge 114+, Safari 17+, Firefox 125+.
* No CSS anchor positioning, Floating UI still does the placement.

Checked in a browser with the sidebar expanded, collapsed, and as a mobile drawer. The panel escapes the rail, sits under its trigger instead of stretching, and keeps its labels.


robsontenorio:
It does not work properly on large screens.
It just float to the left

image

---

Small

image

lauroguedes:
You were right, thanks for catching it.

The trigger was not the screen size, it was where the app starts on screen. Floating UI returns coordinates measured from the nearest containing block, and the sidebar wrapper is one because daisyUI gives it `translate: 0%`. A top layer element is measured from the viewport instead. Those two only agree when the app sits at the top left corner, which is why it looked fine in my app and drifted in yours. I reproduced it by pushing the app 600px to the right and the panel landed exactly 600px off.

So I dropped Floating UI for this mode. The panel is now placed from the trigger's own on screen position, which is already the right coordinate space, plus:

- flips up when there is no room below
- stays inside the viewport, so a panel wider than a collapsed sidebar does not hang off screen (that one was 127px off screen in my own app)
- follows the trigger on scroll and resize, using a capture listener so scrolling any ancestor works too

Checked in a browser with the app pushed away from the corner: sidebar expanded and collapsed, trigger at the bottom of the screen, page scroll, ancestor scroll, resize, and the mobile drawer.

**Nothing changes without the prop.** I rendered 11 dropdown variants without `popover` (default, `right`, `top`, `no-x-anchor` and its combinations, `scroll`, `max-height`, `icon`, `id`, custom trigger) against current `main` and compared the HTML. The only difference is indentation inside the tag. Every attribute, class and value is identical, and `x-anchor` behaves exactly as before.

One deliberate restriction: `popover` combined with `no-x-anchor` now throws. `no-x-anchor` leaves placement to daisyUI, which measures from the `.dropdown` parent, and that parent means nothing once the panel is in the top layer. It failed silently by pinning the panel to the top left corner, so failing loudly seemed better.

PR #1115, opened 2026-08-21
Table: add sortable (drag-and-drop sorting feature) and selectable-condition

amnesca:
**Sortable (drag-and-drop feature):**

Add sortable feature to sort rows by drag-and-drop using Livewire's [wire:sort](https://livewire.laravel.com/docs/4.x/wire-sort). See instructions for documentation below:

**Row drag-and-drop sorting**

Use `sortable` attribute in conjunction with `sortable-method="handleSort"` to sort rows by dragging and dropping them in place. The `sortable-method` is where you provide relevant logic to persist to the database. 

```blade
<x-table ... sortable sortable-method="handleSort">
```
```php
public TodoTable $table;

public function handleSort($id, $position)
{
	$task = $this->table->tasks()->findOrFail($id);

	// Update the task's position and re-order other tasks...
}
```
[Screencast from 2026-08-21 11-22-56.webm](https://github.com/user-attachments/assets/b6ca1683-874e-4347-af80-60ba34c91020)

By default, it will look up for `$row->id` attribute. You can customize this with `sortable-key` attribute.

```blade
{{-- Uses `$row->mycode` as sortable key --}}
<x-table ... sortable sortable-key="mycode" />
```

You can also control the sort icon visibility by using `sortable-condition` attribute.

```blade
{{-- It will check each row for the `is_admin` field --}}
<x-table ... sortable sortable-condition="is_admin" />
```

----------------------------------------------------------------------------------------------------------------------------

**Selectable condition:**

Add condition to `selectable` feature. See instructions for documentation below:

You can also control the select icon visibility by using `selectable-condition` attribute.

```blade
{{-- It will check each row for the `is_admin` field --}}
<x-table ... selectable selectable-condition="is_admin" />
```

### DIFF
--- a/src/View/Components/Table.php
+++ b/src/View/Components/Table.php
@@ -25,9 +25,14 @@ class Table extends Component
         public ?bool $noHeaders = false,
         public ?bool $selectable = false,
         public ?string $selectableKey = 'id',
+        public mixed $selectableCondition = null,
         public ?bool $expandable = false,
         public ?string $expandableKey = 'id',
         public mixed $expandableCondition = null,
+        public ?bool $sortable = false,
+        public ?string $sortableKey = 'id',
+        public mixed $sortableCondition = null,
+        public mixed $sortableMethod = null,
         public ?string $link = null,
         public ?bool $withPagination = false,
         public ?string $perPage = null,
@@ -295,7 +300,11 @@ class Table extends Component
                                 <!-- EXPAND EXTRA HEADER -->
                                 @if($expandable)
                                     <th class="w-1"></th>
-                                 @endif
+                                @endif
+
+                                @if($sortable)
+                                    <th class="w-1"></th>
+                                @endif
 
                                 @foreach($headers as $header)
                                      @php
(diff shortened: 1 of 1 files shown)


robsontenorio:
This will break the current behavior for:

https://mary-ui.com/docs/components/table#row-selection

Because `selectableCondition` is null by default. Or am I wrong?


```
@if(data_get($row, $selectableCondition))

```

amnesca:
Well, I just used the same logic as the existing [expandableCondition](https://github.com/robsontenorio/mary/blob/181757e69074b5a0fc270eceefa1aeff7fb3e958/src/View/Components/Table.php#L358) which works fine, and it seems `data_get($row, null)` returns the row which equates to true in an if statement. Although, perhaps it should use the third default parameter for clarity? ie. `data_get($row, $selectableCondition, true)`.

I also just realised that select all needs to be filtered for `selectableCondition`, so I have amended [getAllIds()](https://github.com/robsontenorio/mary/pull/1115/commits/5731b54d95e7289e61092f57d6dd391a0ab4a1bf) to filter ids of the selectable condition when applied.

Thanks.

